PR #9, opened 2018-07-12
[WIP] Initial form for usecases document

sorinstelian:


sorinstelian:
Will add proper links between sections

PR #10, opened 2018-09-19
Add django-admin-bootstrap theme + django-import-export library

catalinjitea:


ozone/core/admin.py
```diff
@@ -1,66 +1,104 @@
...
-    PartyHistory,
-    ReportingPeriod,
-    Obligation,
-    Annex,
-    Group,
-    Substance,
-    Blend,
-    BlendComponent,
-)
+from .models import (Meeting, Treaty, Decision, Region, Subregion, Party,
```
sorinstelian: Regarding the formatting for these lines, check this out:
https://github.com/ambv/black

I pretty much like having a single line for each item in this case, it makes the import list more readable.

config/settings/base.py
```diff
@@ -72,14 +72,16 @@
...
+    'import_export',
 ]
 LOCAL_APPS = [
     'ozone.users.apps.UsersConfig',
     # Your stuff: custom apps go here
     'ozone.core.apps.CoreConfig',
 ]
 # https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps
-INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
+INSTALLED_APPS = THIRD_PARTY_APPS + LOCAL_APPS + DJANGO_APPS
```
sorinstelian: I suppose these is needed for templates precedence. Is that so?

PR #14, opened 2018-09-24
Update models

catalinjitea:


ozone/core/models/data.py
```diff
@@ -218,6 +218,8 @@ class BaseImportExportReport(BaseBlendCompositionReport, BaseExemption):
         'quantity_feedstock'
     ]
 
+    party = models.ForeignKey(Party, on_delete=models.PROTECT)
```
sorinstelian: No need for a foreign key, `party` is accessible through the `submission` field.

ozone/core/models/party.py
```diff
@@ -165,6 +184,21 @@ def max_value_current_year(value):
     return MaxValueValidator(current_year())(value)
 
 
+class PartyType(models.Model):
```
sorinstelian: Just a thought: perhaps a model is not needed for this at this point; however, I am seeing the value of allowing the OS to add new types through the admin.

sorinstelian: I guess a simple enumeration will do, as the actual business logic would probably rely on the PartyHistory model.

ozone/core/models/party.py
```diff
@@ -205,6 +242,36 @@ class Meta:
         verbose_name_plural = 'parties history'
 
 
+class PartyRatification(models.Model):
+    """
+    Ratification information of all treaties and amendments, for each party.
+    """
+
+    ratification_id = models.CharField(max_length=16, unique=True)
```
sorinstelian: What is this field's purpose?

ozone/core/models/party.py
```diff
@@ -221,3 +288,61 @@ def __str__(self):
...
     class Meta:
         ordering = ('name',)
+
+
+class UsesType(models.Model):
+    """
+    The different categories of uses of controlled substances that need to be reported.
+    """
+
+    uses_type_id = models.CharField(max_length=16, unique=True)
```
sorinstelian: Also not sure what purpose this would serve.

ozone/core/models/reporting.py
```diff
@@ -59,6 +61,30 @@ def __str__(self):
...
+    """
+
+    obligation_type_id = models.CharField(max_length=16, unique=True)
+
+    name = models.CharField(max_length=256, unique=True)
+
+    description = models.CharField(max_length=256, blank=True)
+
+
+class SubmissionStatus(models.Model):
```
sorinstelian: The actual implementation for this will be pretty different, I don't think adding this and then removing it later would server any purpose.

ozone/core/models/data.py
catalinjitea: done.

ozone/core/models/party.py
catalinjitea: done.

ozone/core/models/reporting.py
```diff
@@ -78,6 +104,9 @@ class SubmissionMethods(enum.Enum):
     obligation = models.ForeignKey(
         Obligation, related_name='submissions', on_delete=models.PROTECT
     )
+    obligation_type = models.ForeignKey(
```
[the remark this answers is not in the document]
catalinjitea: I will keep Obligation model and add a description field to this model.

ozone/core/models/reporting.py
catalinjitea: done.

ozone/core/models/party.py
catalinjitea: ratification_id field removed.

PR #17, opened 2018-10-02
Api-nested

catalinjitea:


sorinstelian:
Looks good so far.

sorinstelian:
We do not need to add all serializers in this PR. If you want to add just the current ones for now, let me know. A new PR can be created later with the remaining serializers.

PR #21, opened 2018-10-05
Update models according to Ozone Business Data Tables v2.5

catalinjitea:


ozone/core/models/party.py
```diff
@@ -338,3 +338,189 @@ def __str__(self):
...
+
+    substance = models.ForeignKey(
+        Substance, on_delete=models.PROTECT
+    )
+
+    uses_type = models.ForeignKey(
+        UsesType, on_delete=models.PROTECT
+    )
+
+    cu_category = models.CharField(max_length=256, blank=True)
```
sorinstelian: I would like it if, at least internally, we don't resort to cryptic abbreviations. `critical_use_category` will allow other people looking at the code for the first time to understand it more easily.

ozone/core/models/substance.py
```diff
@@ -213,29 +210,38 @@ class HATProduce(models.Model):
...
 
+    class Meta:
+        abstract = True
+
+
+class HighAmbientTemperatureProduce(BaseHighAmbientTemperature):
+    """
+    Production under the exemption for high-ambient-temperature parties
+    """
+
```
sorinstelian: Empty class definition and no `pass`. :)

ozone/core/migrations/0010_auto_20181008_0915.py
```diff
@@ -0,0 +1,362 @@
...
+            new_name='laboratory_analytical_uses_quantity',
+        ),
+        migrations.RenameField(
+            model_name='article7production',
+            old_name='decision_pau',
+            new_name='process_agent_uses_decision',
+        ),
+        migrations.RenameField(
+            model_name='article7production',
+            old_name='quantity_pau',
```
sorinstelian: Do you think it would be worthwhile to prefix the field names with quantity/decision as they used to be? I believe it makes them easier to read.

ozone/core/models/data.py
```diff
@@ -6,11 +6,13 @@
...
 
 from .meeting import Decision, ExemptionTypes
-from .reporting import Submission
-from .substance import Annex, Group, Substance, Blend, BlendComponent
 from .party import Party
+from .reporting import Submission
+from .substance import BlendComponent, Substance
 
 __all__ = [
+    'BaseReport',
```
sorinstelian: I don't think these two need to be here.

ozone/core/models/data.py
```diff
@@ -160,10 +162,10 @@ class BaseBlendCompositionReport(BlendCompositionMixin, BaseReport):
 
     # `blank=True` is needed for full_clean() calls performed by save()
     substance = models.ForeignKey(
-        Substance, blank=True, null=True, on_delete=models.PROTECT
+        "core.Substance", blank=True, null=True, on_delete=models.PROTECT
```
sorinstelian: do we need to use model names (as strings) here?

ozone/core/models/substance.py
```diff
@@ -3,6 +3,7 @@
 from django.core.validators import MinValueValidator, MaxValueValidator
 from django.db import models
 
+import ozone.core.models.data
```
sorinstelian: I don't think this is needed.

ozone/core/models/substance.py
```diff
@@ -196,3 +199,38 @@ def __str__(self):
 
     class Meta:
         ordering = ('blend', 'substance')
+
+
+class ProcessAgentApplication(models.Model):
```
sorinstelian: This and the following have a better place in the `reporting.py` file.

ozone/core/models/data.py
```diff
@@ -450,9 +452,51 @@ class Article7Emission(BaseReport):
...
+class HighAmbientTemperatureProduce(BaseReport, BaseHighAmbientTemperature):
+    """
+    Production under the exemption for high-ambient-temperature parties
+    """
+    substance = models.ForeignKey(
+        Substance, blank=True, null=True, on_delete=models.PROTECT
+    )
+
+
+class HighAmbientTemperatureImport(BaseBlendCompositionReport, BaseHighAmbientTemperature):
```
sorinstelian: Since this inherits BaseBlendCompositionReport, it should define a tracker field and a list of quantity fields that need to be calculated when substance rows are automatically derived from blend rows.

catalinjitea: Can MSACProd, SDACProd, DCPACProd be considered quantities and should be included in QUANTITY_FIELDS?

sorinstelian: Yes, they are substance/blend quantity fields.

PR #25, opened 2018-10-10
Feature/frontend integration

mihai-macaneata:


Dockerfile
```diff
@@ -8,40 +8,54 @@ ARG BACKEND_PORT=80
...
+    && apt-get install -y --no-install-recommends $runDeps \
+    && apt-get clean \
+    && rm -vrf /var/lib/apt/lists/*
+RUN mkdir -p $APP_HOME
+COPY $REQUIREMENTS_FILE $APP_HOME/
+WORKDIR $APP_HOME
+RUN buildDeps="build-essential gcc" \
+    && apt-get update -y \
+    && apt-get install -y --no-install-recommends $buildDeps \
+    && pip install --no-cache-dir -r $REQUIREMENTS_FILE \
```
sorinstelian: I would keep this in a separate RUN command, as it was before, since system-wide buildDeps and python libraries are not necessarily linked. This way, the layer containing the installed buildDeps can be re-used if python requirements change.

Dockerfile
```diff
@@ -8,40 +8,54 @@ ARG BACKEND_PORT=80
...
+COPY package.json postcss.config.js yarn.lock $APP_HOME/
 WORKDIR $APP_HOME
+RUN npm_config_tmp=$APP_HOME yarn
+RUN npm run build
 
-RUN pip install --no-cache-dir -r $REQUIREMENTS_FILE
-
+FROM python:3.6-slim
+ARG REQUIREMENTS_FILE=requirements.txt
+ENV APP_HOME=/var/local/fise.nfi.search
```
sorinstelian: fise.nfi.search :)

PR #26, opened 2018-10-10
Feature/update models

catalinjitea:


ozone/core/migrations/0014_auto_20181010_1314.py
```diff
@@ -0,0 +1,42 @@
+# Generated by Django 2.0.5 on 2018-10-10 13:14
```
sorinstelian: There is already a migration numbered 13 on the `develop` branch. :) Please fetch and rebase.

sorinstelian: Nevermind. :)

PR #30, opened 2018-10-15
User model

catalinjitea:


config/settings/base.py
```diff
@@ -120,15 +120,14 @@ def get_int_env_var(var_name, default=None):
...
 ]
 
 TOKEN_EXPIRE_INTERVAL = datetime.timedelta(days=get_int_env_var('TOKEN_EXPIRE_INTERVAL'))
 
 
 # https://docs.djangoproject.com/en/dev/ref/settings/#auth-user-model
-AUTH_USER_MODEL = 'users.User'
-# https://docs.djangoproject.com/en/dev/ref/settings/#login-redirect-url
-LOGIN_REDIRECT_URL = 'users:redirect'
+AUTH_USER_MODEL = 'core.User'
```
sorinstelian: Not a change for this file, but all models containing foreign keys to User should instead reference `settings.AUTH_USER_MODEL` (or use `get_user_model()`)

ozone/users/admin.py
```diff
@@ -1,39 +0,0 @@
-from django import forms
```
sorinstelian: Since we are not registering this anymore, we should register our custom user model in the admin.

PR #34, opened 2018-10-18
Implement update-prevention of data report models and submissions

catalinjitea:


ozone/core/models/data.py
```diff
@@ -316,7 +331,8 @@ class Meta:
 
 
 class Article7Export(
-    BaseBlendCompositionReport, BaseImportExportReport, BaseUses
+    ModifyPreventionMixin, BaseBlendCompositionReport, BaseImportExportReport,
```
sorinstelian: Both `ModifyPreventionMixin` and `BaseBlendCompositionReport` alter the behavior of clean(). This will not work.

sorinstelian: Nevermind, my bad.

ozone/core/models/reporting.py
```diff
@@ -355,6 +355,20 @@ class Meta:
...
+    def delete(self, using=None, keep_parents=False):
+        if not self.data_changes_allowed:
+            raise Exception(
+                _("Submitted submissions cannot be modified.")
+            )
+        super().delete()
+
+    def clean(self):
+        if not self.data_changes_allowed:
+            raise Exception(
```
sorinstelian: I think we should raise `ValidationError`s in `clean()` and `RuntimeError`s in `save()` or `delete()`

PR #60, opened 2018-10-25
Update BlendCompositionMixin save method

catalinjitea:


ozone/core/models/data.py
```diff
@@ -140,15 +140,13 @@ def save(self, *args, **kwargs):
...
-                report = self.__class__.objects.create(
-                    submission=self.submission,
-                    substance=component.substance,
-                    blend=None,
-                    blend_item=self,
-                    # Quantities for specific substance
-                    **field_dictionary
-                )
+                report = self.__class__.objects.get(pk=self.pk)
+                report.pk = None
```
sorinstelian: Nice and elegant solution, but I would add a comment regarding what this line does
(https://docs.djangoproject.com/en/2.1/topics/db/queries/#copying-model-instances)

PR #63, opened 2018-10-26
Add blend and blend component fixtures

catalinjitea:


sorinstelian:
These should also be added to the load_initial_fixtures mgmt command, with care regarding the order of loading. :)

PR #90, opened 2018-10-30
Resolves #68

### LINKED ISSUE #68: Models - Submission - transitioning to "Submitted" should make submission current
When a submission becomes "Submitted" (from "data entry") or when it is un-recalled, it should become "current".

N.B.: in the case of an accelerated workflow (only available for Secretariat), "data entry" goes directly to "finalized", bypassing the "Submitted" state. This should also make the transition current. While this does not have to be implemented right now, we should take care that the implementation is generic enough to make this easy.

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -397,6 +397,16 @@ def save(self, *args, **kwargs):
         self.clean()
         return super().save(*args, **kwargs)
 
+    def make_current(self):
```
sorinstelian: We should also set the superseded flag to False for the current submission.

ozone/core/models/reporting.py
```diff
@@ -397,6 +397,16 @@ def save(self, *args, **kwargs):
...
+        versions = Submission.objects.filter(
```
sorinstelian: I think a `select_for_update()` is a safer choice here.

PR #95, opened 2018-11-01
Resolves #44 Add endpoint to clone a submission

### LINKED ISSUE #44: Api for cloning a submission
The cloning should include all data in the submission.
The version and state should change

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -325,6 +327,27 @@ def non_exempted_fields_modified(self):
...
+        """
+        Checks whether the current submission can be cloned
+        """
+        current_date = datetime.now().date()
+        current_submissions = Submission.objects.filter(
+            reporting_period__start_date__lte=current_date,
+            reporting_period__end_date__gte=current_date,
+            obligation=self.obligation
+        )
+        if self in current_submissions:
```
sorinstelian: I do not understand these conditions, we should talk about them. :)

ozone/core/api/views.py
```diff
@@ -161,6 +162,20 @@ def list(self, request, *args, **kwargs):
         )
         return Response(serializer.data)
 
+    @action(detail=True, methods=['post'])
+    def clone(self, request, pk=None):
```
sorinstelian: My bad for not thinking this through properly, but this will not duplicate the data related to the submission, which is actually the real purpose of "cloning" submissions.

We should also make sure that we set the proper `created_by` field on the submission

And finally, I believe that most of the code in this method should be in the Submission model rather than here, in the view.

PR #110, opened 2018-11-16
Resolve #50 Change history for each submission

### LINKED ISSUE #50: API: Change history for each submission
This endpoint will list all changes made in the submission.
For each change made in the submission will contain:
- Date
- Event ( CREATE, ADD, CHANGE, SUBMIT, PROCESS, RECALL etc.)
- User 
- State

catalinjitea:


sorinstelian:
We also have the `TransitionEvent` model which implements a part of this functionality. It should be removed, as the implementation in this PR is more general and includes what we have there.

ozone/core/models/reporting.py
```diff
@@ -173,6 +173,8 @@ class TransitionNotAvailable(Exception):
     # Needed to track state changes and help with custom logic
     tracker = FieldTracker()
 
+    log_entry = AuditlogHistoryField()
```
sorinstelian: I guess this also requires a migration. :)

catalinjitea: We don't need a migration for this field and I think, the reason for this is that the functionality is built on Django’s content types framework, more exactly using GenericRelation.

ozone/core/api/urls.py
```diff
@@ -141,6 +141,11 @@ def extend(self, added_router):
             'submission-versions/',
             views.SubmissionVersionsListViewSet.as_view(),
             name='submission_versions',
+        ),
+        re_path(
```
sorinstelian: Instead of this approach, we can implement a `history` method in the ViewSet and mark it with `@action(detail=True, methods=['get'])` - see https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

Issue #126, opened 2018-11-20
BlendComponents without substance id


melish:
Related to #106

sorinstelian:
Fully solved in 7fb080a0a62c4e3a6f0af8be71a4fa4ca7b4efd2

Issue #125, opened 2018-11-20
Other Substances and nullable foreign keys to Annex/Grp


melish:

sorinstelian:
Decision was made to make Annex/Group null-able in the Substance model. If they are null, it means that the substance is not controlled.

Also, Substance should only have a FK to Group at this point, since Annex information can be taken from there.

@melish : please check this is correct. :)

sorinstelian:
Implemented as described above.

Issue #122, opened 2018-11-20
New fields in BlendComponent


melish:
component_name, CNumber - to be discussed

Issue #121, opened 2018-11-20
Add sort order to Substance model


melish:
From column AnxGrpSort in the Excel file

And also sort list of substances in data entry form

PR #127, opened 2018-11-20
Command to generate fixtures from xlsx + small change in substance model

melish:
Delete+reload fixtures needed
Also related to #120, #121, #122, #123, #125, #126


ozone/core/models/substance.py
```diff
@@ -196,7 +196,7 @@ class BlendComponent(models.Model):
     )
 
     substance = models.ForeignKey(
-        Substance, related_name='blends', on_delete=models.PROTECT
```
sorinstelian: Wouldn't it be better to use a `is_controlled` flag on the Substance model? (though that complicates things a bit)

Or it may happen that the users don't want to report non-controlled substances in their blends? :)

PR #139, opened 2018-11-22
Feature/submission history

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -175,6 +174,8 @@ class TransitionNotAvailable(Exception):
     # Needed to track state changes and help with custom logic
     tracker = FieldTracker()
 
+    history = HistoricalRecords()
+
```
sorinstelian: https://django-simple-history.readthedocs.io/en/latest/advanced.html#using-history-user-to-record-which-user-changed-a-model - should't this be done here as well? does the implementation in this PR work correctly with multiple users?

ozone/core/serializers.py
```diff
@@ -512,6 +512,31 @@ class Meta(CreateSubmissionSerializer.Meta):
...
+        return obj.history_date.strftime('%Y-%m-%d')
+
+    def get_current_state(self, obj):
+        # Unfortunately I can't find a way to avoid using the protected field
+        return obj._current_state
+
+    def get_user(self, obj):
+        if obj.history_user:
+            return obj.history_user.username
+        return None
```
sorinstelian: When does this happen? :)

catalinjitea: Yep, you're right.

ozone/core/migrations/0012_historicalsubmission.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 2.0.5 on 2018-11-22 12:53
```
sorinstelian: please update migration

catalinjitea: Done.

PR #153, opened 2018-11-26
After cloning a submission return id; proper error messages added also

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -374,20 +374,28 @@ def check_cloning(self):
...
         ):
-            return True
-        else:
-            return False
+            if not (
+                self.current_state == "finalized"
+                and self.flag_valid is True
+                and self.flag_superseded is False
+            ):
+                raise ValidationError(
```
sorinstelian: Why do we raise ValidationErrors in a method that returns True/False?

PR #155, opened 2018-11-26
Move validation logic in model

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -385,14 +385,21 @@ def check_cloning(self):
...
+            try:
+                clone = Submission.objects.create(
+                    party=self.party,
+                    reporting_period=self.reporting_period,
+                    obligation=self.obligation,
+                    cloned_from=self,
+                    created_by=self.created_by,
+                    last_edited_by=self.last_edited_by
+                )
+            except ValidationError as e:
```
sorinstelian: why is this try-except block needed if the error is simply re-raised?

PR #213, opened 2018-12-03
Resolves #192 RESTful error messages handling

### LINKED ISSUE #192: Error messages
Look for all 500 internal server errors and return a response containing 4** status code.

catalinjitea:


ozone/core/models/data.py
```diff
@@ -12,6 +11,7 @@
 from .reporting import Submission
 from .substance import BlendComponent, Substance, Blend, Annex, Group
 from .utils import model_to_dict
+from ozone.core.exceptions import CustomValidationError
```
sorinstelian: Let's stick to the same style of imports (relative vs aboslute) - it both keeps style consistent and makes it easier to refactor code in the future (e.g. changing  app name to something else than 'core')

ozone/core/models/data.py
```diff
@@ -36,8 +36,8 @@ class ModifyPreventionMixin:
 
     def clean(self):
         if not self.submission.data_changes_allowed:
-            raise ValidationError(
-                _("Submitted submissions cannot be modified.")
+            raise CustomValidationError(
+                "Submitted submissions cannot be modified."
```
sorinstelian: why lose the `gettext_lazy`? :)

ozone/core/models/reporting.py
```diff
@@ -472,7 +474,7 @@ class Meta:
 
     def delete(self, *args, **kwargs):
         if not self.data_changes_allowed:
-            raise RuntimeError(
+            raise CustomValidationError(
```
sorinstelian: There is no validation performed here, I suggest using a different error & status code in this case (e.g. method not allowed).

PR #218, opened 2018-12-04
Feature/simple permissions

sorinstelian:


ozone/core/migrations/0026_user_is_read_only.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.0.5 on 2018-11-29 12:50
```
sorinstelian: Need to update this migration. :)

PR #224, opened 2018-12-07
Refactor endpoint to return list of non-parties per group

### LINKED ISSUE #207: Filter list of parties in tab Non party trade
using the new API for non-parties 

Note: Islands (territories dependent on a non-party) should remain included in the list.

catalinjitea:


ozone/core/api/views.py
```diff
@@ -163,14 +163,18 @@ def get_queryset(self):
...
-            self.kwargs["substance_pk"]
-        )
+    renderer_classes = (JSONRenderer, )
+
+    def get(self, request):
+        groups = Group.objects.all()
+        all_non_parties = {}
+        for group in groups:
+            queryset = Article7NonPartyTrade.get_non_parties(group.pk)
+            non_parties_per_group = {key.id:True for key in queryset.all()}
```
sorinstelian: I would use something like `{id: True for id in queryset.values_list('id', flat=True)}` so we only fetch the id's from the database and don't cache stuff needlessly in the ORM

Issue #234, opened 2018-12-10
Refactor PartyTypes to be a db model instead of enum


melish:
Also, add column: is_a5 (True/False).
Related to #233

sorinstelian:
Solved in https://github.com/eaudeweb/ozone/pull/243

PR #243, opened 2018-12-11
Making PartyType model instead of enum - solves #234

sorinstelian:
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/243)


melish:
Looks good, but for some reason is_article5 is not present in the fixtures

sorinstelian:
Re-generated the PartyHistory fixtures to include is_article5 flag

melish:
I have this error when running the migration: 
```
psycopg2.DataError: invalid input syntax for integer: "Non Article 5"
```

Issue #226, opened 2018-12-07
Take user permissions into account at state change and cloning (including is_cloning and available states)


sorinstelian:

melish:
Is this issue stuck and you need my support?

sorinstelian:
Not stuck, Catalin was waiting for specs from me on who can perform transitions, working on it.

sorinstelian:
### Transitions - roles mapping (who can perform which transitions):

**Data Entry - Submitted**
- Party-write user from Party to which the submission refers to, *only if* the submission was started by the Party
- Secretariat-write user, *only if* the submission was started by Secretariat

**Any state - Recalled**
- Secretariat-write user, *only if* the submission was started by Secretariat
- Party-write user from Party to which the submission refers to, *only if* the submission was started by the Party

**Recalled - Any state**
- Secretariat-write user, *only if* the submission was started by Secretariat
- Party-write user from Party to which the submission refers to, *only if* the submission was started by the Party

**Submitted - In Processing**
- Secretariat write user

**In Processing - Finalized**
- Secretariat write user

PR #307, opened 2018-12-13
267 migrate art7 reports

alexkiro:
Closes #267

---
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/307)

PR #309, opened 2018-12-13
Take user permissions into account at state change, refs #226

catalinjitea:
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/309)

### DIFF
--- a/ozone/core/models/workflows/accelerated.py
+++ b/ozone/core/models/workflows/accelerated.py
@@ -34,6 +34,18 @@ class AcceleratedArticle7Workflow(BaseWorkflow):
 
     state = AcceleratedArticle7WorkflowStateDescription()
 
+    @xworkflows.transition_check('finalize')
+    def check_finalize(self):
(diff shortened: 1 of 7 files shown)


catalinjitea:
Connect #226

sorinstelian:
Let's also do the needed changes to the import script:
https://github.com/eaudeweb/ozone/pull/307#discussion_r241322854

ozone/core/models/workflows/accelerated.py
```diff
@@ -34,6 +34,18 @@ class AcceleratedArticle7Workflow(BaseWorkflow):
 
     state = AcceleratedArticle7WorkflowStateDescription()
 
+    @xworkflows.transition_check('finalize')
```
sorinstelian: what happens here if user is None?

catalinjitea: I was trying to avoid the extra verification because this check needs to be done in every 'transition check' function. So, when we call transition function, we need to make sure that we pass the `User`.

PR #327, opened 2018-12-14
Tests for workflow permissions

catalinjitea:
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/327)

### DIFF
--- a/ozone/tests/test_clone_permissions.py
+++ b/ozone/tests/test_clone_permissions.py
@@ -1,0 +1,138 @@
+from django.urls import reverse
+from django.test import TestCase
+from django.contrib.auth import get_user_model
+from django.contrib.auth.hashers import Argon2PasswordHasher
+
+from .factories import (
+    PartyFactory,
+    RegionFactory,
+    ReporterUserFactory,
+    SecretariatUserFactory,
+    SubmissionFactory,
+    SubregionFactory,
+)
+
+
+User = get_user_model()
+
+
+class ClonePermissionsTests(TestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.hash_alg = Argon2PasswordHasher()
+        self.region = RegionFactory.create()
+        self.subregion = SubregionFactory.create(region=self.region)
+
+    def get_authorization_header(self, username, password):
+        resp = self.client.post(reverse("core:auth-token-list"), {
+            "username": username,
+            "password": password,
+        }, format="json")
+        return {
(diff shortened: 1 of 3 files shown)


ozone/tests/test_clone_permissions.py
```diff
@@ -0,0 +1,138 @@
...
+        headers = self.get_authorization_header(reporter_same_party.username, 'qwe123qwe')
+        resp = self.client.post(
+            reverse("core:submission-clone",
+                    kwargs={'pk': submission.pk}),
+            {"party": party.pk},
+            **headers
+        )
+        self.assertEqual(resp.status_code, 200)
+
+    def test_clone_another_party(self):
```
sorinstelian: maybe we should also do a similar test with submissions created by the secretariat?

ozone/tests/test_workflow_permissions.py
```diff
@@ -0,0 +1,580 @@
...
+        resp = self.call_transition(
+            owner=self.secretariat_user,
+            user=secretariat_user_ro,
+            party=party,
+            current_state='finalized',
+            transition='recall'
+        )
+        self.assertEqual(resp.status_code, 403)
+
+    def test_unrecall_non_ro_secretariat(self):
```
sorinstelian: let's use `write` instead of `non_ro` :)

ozone/tests/test_workflow_permissions.py
```diff
@@ -0,0 +1,580 @@
...
+    def test_finalize_ro_secretariat(self):
+        """
+        Testing `finalize` transition using a secretariat user with read only
+        permissions.
+        Expected result: 403 Forbidden.
+        """
+
+        party = PartyFactory(subregion=self.subregion)
+        secretariat_user_ro = SecretariatUserFactory(
+            username='secretariat_user_ro',
```
sorinstelian: I think we should look into refactoring this a bit: things like username and email are in no way relevant to the test and could be hidden away in factories.

ozone/tests/test_workflow_permissions.py
```diff
@@ -0,0 +1,580 @@
...
+            email='secretariat_user_ro@example.com',
+            password=self.hash_alg.encode(password='qwe123qwe', salt='123salt123'),
+            is_read_only=True
+        )
+        resp = self.call_transition(
```
sorinstelian: Also, it's not very obvious that `call_transition` actually does two things: creates a submission and then tries to trigger a transition on it. I think it should be broken down in two distinct methods.

PR #328, opened 2018-12-14
322 remove unk and czs

### LINKED ISSUE #322: Remove Unspecified and Czechoslovakia from fixtures
These two parties should be removed, but addressing the implications:
- [x] the pk's of remaining parties will probably have to change
- [x] existing initial/test data (fixtures, submissions, etc.) should be smoothly migrated
- [x] legacy data imported in #318 and #313 will have to be executed again to double-check if there is anything related to these two invalid countries

alexkiro:
Closes #322

---
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/328)


alexkiro:
Travis build failed, need to fix that before we can merge this.

alexkiro:
> Travis build failed, need to fix that before we can merge this.

Error was because the migration was attempting to refresh only the changed fixtures in a clean environment, and hence not having all the data for all the relationships. 

Fixed, by not running the refresh on clean envs.

PR #343, opened 2018-12-19
332 use backend pagination/sorting/filtering on the dashboard

### LINKED ISSUE #332: Add backend pagination/sorting/filters to the dashboard
At the moment the dashboard will take a long time to load if there are many results as the pagination and sorting is done in the frontend.

We should:

- [x] Add pagination/sorting support in the backend for the submission list API
- [x] Use the backend pagination/sorting support in the dashboard's frontend
- [x] Use a default sort order (by last modified desc, then by reporting period desc, then by party name, then by reporting obligation - art 7 first)
- [x] _(Nice to have)_ Also move the dashboard filters to the backend instead of the frontend

alexkiro:
Closes #332 

Has both backend and frontend implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/eaudeweb/ozone/343)


ozone/core/api/views.py
```diff
@@ -251,23 +252,72 @@ class UserViewSet(ReadOnlyMixin, viewsets.ModelViewSet):
...
+class SubmissionPaginator(PageNumberPagination):
+    page_size = 10
+    page_query_param = 'page'
+    page_size_query_param = "page_size"
+
+
+class SubmissionViewFilterSet(filters.FilterSet):
+    party = filters.NumberFilter("party", help_text="Filter by party ID")
+    obligation = filters.NumberFilter("obligation", help_text="Filter by Obligation ID")
+    reporting_period = filters.NumberFilter("reporting_period",
```
sorinstelian: We have tried to use the code formatting style described [here](https://black.readthedocs.io/en/stable/the_black_code_style.html): throughout the project. I think it would be good to stay consistent at this point, since there is a lot of code already written. :)

alexkiro: Reformatted with the tool. :+1: 

FWIW, I only formatted the classes that I've changed/added here. But running the tool on the whole project does change a lot more files. 

@sorinstelian maybe it's worth doing a run on the whole project?

sorinstelian: @alexkiro : yup, it's definitely worth it; perhaps it would be better to defer that though for when we get back from the holidays so we don't mess anything up? :)

ozone/core/api/views.py
```diff
@@ -118,9 +119,9 @@ def filter_queryset(self, request, queryset, view):
             return queryset
         else:
             # Party user
-            if queryset and queryset.model == Submission:
+            if queryset is not None and queryset.model == Submission:
```
sorinstelian: @alexkiro : good catch! 👍

ozone/core/api/views.py
alexkiro: :laughing: agreed !

Issue #374, opened 2018-12-20
R/O vs R/W for Remarks and Comments fields


melish:
Permissions should be established for editing the party/secretariat remarks, and new general remarks added for each data form.

* [x] Parties should view the Remarks (secretariat) and Comments (secretariat), but not edit them
* [x] If the submission was created by a party user, then the "Party remarks" and "Party comments" should remain read-only for Secretariat users
* [x] New per-submission fields for remarks_secretariat, and remarks_party should be added for each data form (import/exports/etc.)
* [x] The restrictions should be applied to both the new fields, and the existing fields per entry in each data form (import/exports/etc.)
* [x] The new general party remarks should only be modifiable while the submission is in `data_entry`
* [x] The new general submission remarks can be modified regardless of the current state of the submission.

sorinstelian:
@mihai-macaneata : for this I think I need to properly return the `filled_by_secretariat` field on the submission detail endpoint and also give you a user detail endpoint.

sorinstelian:
On the backend side, general comments/remarks fields will need to be added on the Submission model for each of the data forms.

melish:
@sorinstelian Parties may need to amend a comment entered by the Secretariat without changing the data. If we make them read only, they will have to create a new version. I prefer to leave as it is and double check with the Secretariat before making a decision.

PR #385, opened 2018-12-21
Adding current-user endpoint; making filled_by_secretariat property on Submission

sorinstelian:
Related  to #374


sorinstelian:
@alexkiro : it is removed :) , check out ozone/core/management/commands/import_submissions.py

PR #408, opened 2019-01-04
374 permissions remarks

mihai-macaneata:


alexkiro:
> However there is an MixIn for in this in drf-extensions, which does the same thing. Will look into that and see if we can use it instead.

I removed our custom code, and switched to `drf-extensions` in 33ab9a1

ozone/core/models/reporting.py
```diff
@@ -428,6 +470,20 @@ def get_exempted_fields():
             "flag_has_reported_c3",
             "flag_has_reported_e",
             "flag_has_reported_f",
+            # Remarks, secretariat remarks can be change
```
sorinstelian: Is that in the specs? :)

Issue #229, opened 2018-12-10
Data form 8 (HAT obligation)


melish:

nico4:
Countries reporting under HAT obligation will be presented a webform similar with Article 7 webform consisting in 4 tabs:

1. Submission info (exactly the same as for Article 17)
2. Data form 7 (related with #228  )
3. Data form 8 (scope of this issue)
4. Attachments (exactly the same as for Article 17)

## Initial specification
<img width="1038" alt="screenshot 2018-12-11 at 15 31 28" src="https://user-images.githubusercontent.com/533311/49803898-da3abf80-fd59-11e8-87d4-21ae6733ceb9.png">

## Weborm specifications

Tab menu name: Production

From left to right the elements of the webform are:

i) Fill in this form....
h1. Production
Subtitle. Annex F substances for exempted subsectors, \n in metric tonnes (not ODP or CO2-equivalent tonnes)

### Main table

1. Search on all columns
2. 5 columns:

- (1) Annex/group (rowspan: 2)
- (2) Substance (rowspan: 2)
- Quantity of new substances produced for approved subsectors to which the high-ambient-temperature exemption applies (production should be for use within the producing country (i) [For each substance produced for use in subsectors that may be approved after the assessments under paragraphs 32 and 33 of decision XXVIII/2, please specify the approved subsector. Should the column space be insufficient, further information can be provided in the “comments” box above.] (colspan: 3):
  - (3) New production for use in multi-split air conditioners
  - (4) New production for use in split ducted air conditioners
  - (5) New production for use in ducted commercial packaged (self-contained) air conditioners

### Other elements

Comments (party or secretariat depending on the user rights)

alexkiro:
Backend API for HAT production completed and pushed into `develop`

PR #415, opened 2019-01-07
229 hat production API

alexkiro:
Connects #229

*Backend changes only* for HAT production API


ozone/core/models/reporting.py
```diff
@@ -229,6 +230,9 @@ class Submission(models.Model):
     nonparty_remarks_secretariat = models.CharField(max_length=9999, blank=True)
     emissions_remarks_party = models.CharField(max_length=9999, blank=True)
     emissions_remarks_secretariat = models.CharField(max_length=9999, blank=True)
+    # HAT Remarks
+    hat_production_remarks_party = models.CharField(max_length=9999, blank=True)
```
sorinstelian: Shouldn't one of these be also added in exempted_fields?

alexkiro: Nice catch :) indeed the secretariat one should be added there to be consistent with the other fields. Done in 1564d8e9.

alexkiro: FWIW, there's an assumption in the tests, that all of them work the same. Will look removing this and checking all fields.

```python
        # XXX Assume this works correctly for all other fields.
        field = "imports_remarks_%s" % field
```

alexkiro: > Will look removing this and checking all fields.

Done in 7366f9617f

ozone/core/migrations/0051_merge_20190107_1623.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.1.4 on 2019-01-07 13:23
```
sorinstelian: Wouldn't it be better to re-create the previous migration instead of adding a new one?

sorinstelian:
Looks good except for the extra migration.

Issue #437, opened 2019-01-09
Permission issues when saving data form 


nico4:
https://cl.ly/6e2450d36ac3

As a Party I'm trying to save the data form on the staging server. The error is: You do not have permission to perform this action.

PR #456, opened 2019-01-11
447 Editable remarks on each data entry after submit 

catalinjitea:


ozone/core/api/views.py
```diff
@@ -496,7 +496,7 @@ class SubmissionRemarksViewSet(mixins.UpdateModelMixin, mixins.ListModelMixin,
     Update the general remarks for this specific submission.
     """
     serializer_class = SubmissionRemarksSerializer
-    permission_classes = (IsAuthenticated, IsSecretariatOrSamePartySubmission)
+    permission_classes = (IsAuthenticated, IsSecretariatOrSamePartySubmissionRelated)
```
alexkiro: FWIW, I fixed this on develop as well. See #437 , might cause a conflict?

alexkiro:
FWIW, I would suggests adding one or two tests for this in `test_submission_data_remarks`. But could be done after merging. :)

PR #462, opened 2019-01-14
451 Other obligations

catalinjitea:


ozone/core/migrations/0036_obligations.py
```diff
@@ -1,13 +1,60 @@
 from django.db import migrations
 from django.core.management import call_command
 
+obligation_data = [
```
sorinstelian: Err why does this modify a previous migration?

alexkiro:
Should we adjust the `mkfixtures` management command to include these? Just in case we need to re-run the script with a new Excel file?

ozone/core/migrations/0036_obligations.py
catalinjitea: Migration 0036_obligations.py calls `loaddata` command on obligation.json fixture, which was modified in this PR to introduce _other_ obligations. So I decided to save the fixture (before the modifications) in a variable to avoid an error.

alexkiro:
> Should we adjust the `mkfixtures` management command to include these? Just in case we need to re-run the script with a new Excel file?

Disregard this :smile: , obligations are not generated with this script, so not relevant.

sorinstelian:
> Should we adjust the `mkfixtures` management command to include these? Just in case we need to re-run the script with a new Excel file?

Not for now at least, since obligations are not included in the fixtures we are receiving.

ozone/core/migrations/0036_obligations.py
sorinstelian: Is there an error to be avoided? :)

catalinjitea: Yes. If you delete all migrations and run them again, at 0036_obligations you will receive:
```
column "other" of relation "core_obligation" does not exist
```

sorinstelian: Fair enough.

sorinstelian:
@catalinjitea : backend tests seem to fail at `create_submission`

catalinjitea:
> @catalinjitea : backend tests seem to fail at `create_submission`

Working on it.

PR #464, opened 2019-01-15
460 clone history

### LINKED ISSUE #460: Version disappears after submission is cloned.
After cloning a submission only the current version is displayed, the old versions are missing. 

https://cl.ly/740f53aa4a4e

Also, please remove Search input, we don't need filters here and found a way to highlight the current version.

alexkiro:
Closes #460.


alexkiro:
Should we adjust the `mkfixtures` management command to include these? Just in case we need to re-run the script with a new Excel file?

Issue #429, opened 2019-01-08
Reporting channel should only be editable for secretariat users


sorinstelian:
But only if the Submission has been created by OS.

Also, even if in the interface it should remain in the Submission Info tab, in the backend it should be moved in the Submission model

mihai-macaneata:
This will be solved by #421

sorinstelian:
Frontend part of this will be solved by #421, backend will need its own code.

catalinjitea:
@sorinstelian can we close this issue?

sorinstelian:
@catalinjitea Not until #429 is solved, I guess.

sorinstelian:
Closing, as #421 has been closed.

PR #472, opened 2019-01-15
Move reporting channel in the Submission model and do the refactoring…

catalinjitea:
…, refs #429


ozone/core/serializers.py
```diff
@@ -786,6 +786,35 @@ class Meta:
         exclude = ('submission',)
 
 
+class UpdateSubmissionInfoAndReportingChannelSerializer(serializers.ModelSerializer):
```
sorinstelian: I think it's strange to use a SubmissionInfo serializer to change fields on the Submission model.

I also think this serializer is not needed at all. Since we pretty much cannot avoid doing the reporting channel check in the serializer, it would be easier to override `is_valid` in the Submission serializer

ozone/core/api/views.py
```diff
@@ -461,11 +462,25 @@ def put(self, request, *args, **kwargs):
         info = Submission.objects.get(pk=self.kwargs['submission_pk']).info
         reporting_channel_name = request.data.get('reporting_channel')
         if reporting_channel_name:
-            reporting_channel_id = ReportingChannel.objects.get(
-                name=reporting_channel_name
-            ).pk
-            request.data['reporting_channel'] = reporting_channel_id
-        serializer = UpdateSubmissionInfoSerializer(info, data=request.data)
+            try:
```
sorinstelian: Do we really need to perform this check here? Why not let it fail in the model?

sorinstelian: Having thought about it more, I still don't think there's any need for this check here. The added value of the error message is almost non-existent, as the UI should, in most cases, send a proper channel here. This is not something that should normally reach the user except for some weird corner cases - not worth the extra code.

PR #489, opened 2019-01-17
436 tests for tusd

alexkiro:
Adds some basic tests for the backend hooks for TUSD.

Only two simple test cases added for success scenario party/secretariat. TODOs added for some edge cases.

### DIFF
--- a/ozone/core/api/views.py
+++ b/ozone/core/api/views.py
@@ -886,7 +886,7 @@ class UploadHookViewSet(viewsets.ViewSet):
             file_info_path.unlink()
 
         except UploadToken.DoesNotExist:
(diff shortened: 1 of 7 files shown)


ozone/core/api/views.py
```diff
@@ -886,7 +886,7 @@ def handle_post_finish(request):
             file_info_path.unlink()
 
         except UploadToken.DoesNotExist:
-            log.error(f'UPLOAD denied for "{token.user}": INVALID TOKEN')
+            log.error(f'UPLOAD denied for "{tok}": INVALID TOKEN')
```
sorinstelian: :+1:

PR #493, opened 2019-01-17
436 downloading attached files

### LINKED ISSUE #436: Implement endpoint for downloading attached files

alexkiro:
Closes #436


ozone/tests/factories.py
```diff
@@ -28,7 +28,7 @@
     HighAmbientTemperatureProduction,
     HighAmbientTemperatureImport,
     UploadToken,
-)
+    SubmissionFile)
```
sorinstelian: Really nitpicking, but let's keep a consistent style. :)

sorinstelian:
Looks good; however tests seem to be failing.

ozone/tests/factories.py
alexkiro: Fixed! :)

alexkiro:
> Looks good; however tests seem to be failing.

Indeed, missed that those started failing after I added support for non-ascii files names. They ought to be passing now. I'll wait for Travis to finish his checks and will merge afterwards.

Issue #418, opened 2019-01-07
Available data reports on a submission should be based on submission type


sorinstelian:
Right now HAT reports could theoretically be added through the API on an Article 7 submission.

sorinstelian:
The questionnaire should also be made available/unavailable based on the obligation.

PR #507, opened 2019-01-18
418 obligation data consistency

alexkiro:
Refs #418 

Should be done, but still need to verify it works as expected and add some tests.


ozone/core/api/views.py
```diff
@@ -712,9 +726,10 @@ def perform_create(self, serializer):
 
 
 class DataOtherViewSet(SerializerDataContextMixIn, viewsets.ModelViewSet):
+    form_types = None
```
catalinjitea: Shouldn't form_types be `other` in this case?

alexkiro: Nice catch :smile: , should be fixed in d63c8ee3d468

alexkiro:
Should be finished now, together with test to check that it actually works.

sorinstelian:
Reviewing :)

PR #526, opened 2019-01-22
517 reporting channel bug

### LINKED ISSUE #517: Saving /api/submissions/pk/submission-info fails on staging
User: party
with error : __all__: ["Unexpected Error!"]
Can't reproduce the bug locally. 
Maybe someone check the django logs on staging ?

catalinjitea:
closes #517.

### DIFF
--- a/ozone/core/management/commands/add_reporting_channel.py
+++ b/ozone/core/management/commands/add_reporting_channel.py
@@ -1,0 +1,21 @@
+"""Quick fix for staging where reporting channel has been lost for all submission.
+Adds reporting channel 'Web form' to all submissions that don't have a
+a reporting channel and submissions are all in data entry.
+"""
+
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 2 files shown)


ozone/core/management/commands/add_reporting_channel.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.core.management.base import BaseCommand
+
+from ozone.core.models.reporting import Submission, ReportingChannel
+
+
+class Command(BaseCommand):
+    help = __doc__
+
+    def handle(self, *args, **options):
```
sorinstelian: Efficiency is not that important in this case, but the best way to do such an operation would be something like:
```
        reporting_channel = ReportingChannel.objects.get(name='Web form')
        Submission.objects.filter(
            reporting_channel__isnull=True,
            _current_state='data_entry'
        ).update(reporting_channel=reporting_channel)
```

Issue #98, opened 2018-11-08
Implement resumable upload endpoint for attachments


sorinstelian:

sorinstelian:
Implemented in https://github.com/eaudeweb/ozone/pull/433.

Uploaded files can be seen in API using api/submission/<id>/files

danmisailescu:
Implemented frontend upload component with progressbar
![Untitled.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5c0fcbafc51ee00011888144/125516c66e82c728ace21e0d46dd9a8e2a968fac8babf301b44fff9e2c4533317346f6282083ba2eb9447f04565b4bf941595c10fbacdf7fa0e32e6c981e5ab38532022d8c6853beb2901eb80bbb258f4885b0bf8348c0ef3d25f5c6.png)

In this.$store.state.form.tabs[this.tab.name].form_fields.attachments 
sunt salvate obiectele JSON cu structura de mai jos:
![Untitled1.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5c0fcbafc51ee00011888144/125516c66e82c728ace21e0d46dd9a8e2a968fac8bb8fd0fe64de29b234324676b5ee17e73d5e571b9447d050a564aa914585e47a8fedc7df2b06962964e49bc8f3900299f7744eeaa880dbe08b0218e4b82b6ba8e4ec7e93c25e399c2.png)

Ca sa termin task-ul mai am nevoie de doua lucruri:
1)  De unde iau Date uploaded? Ar trebui sa vina de pe serverul de TUS la fel cum vine si  uploadUrl. Nu putem avea incredere in clientside ca sa setam date uploaded

2)  La Save and Continue pe forma sa salvam si acel array cu attachments de mai sus. Structura JSON-urilor este corecta? Ce format asteapta API-ul? La ce URL?  

@sorinstelian @mihai-macaneata @catalinjitea

sorinstelian:
At this point the upload flow seems to work fine from frontend to backend.
@danmisailescu :
- files uploaded to a submission can be deleted with a DELETE request to /submissions/<submission_id>/files/<file_id>/ (file_id is the id of the file as found in the list returned by a GET to ../files/)
- right now, uploading a file with an already-existing name to the same submission will completely overwrite the new file. @nico4 : is this OK? Should we at least give a warning?
- files can be downloaded using a GET to /submisisons/<submission_id>/files/<file_id>/download/

PR #529, opened 2019-01-22
#98 integration work for uploaded files status + debugging and commented…

danmisailescu:
… breaking python changed that rejected all files with valid extensions with error: bad file extension

### DIFF
--- a/ozone/core/api/views.py
+++ b/ozone/core/api/views.py
@@ -828,16 +828,20 @@ class UploadHookViewSet(viewsets.ViewSet):
                     {'error': 'filename not in MetaData'},
                     status=status.HTTP_400_BAD_REQUEST
                 )
+            """
+            TODO: Fix this code
 
-            if not SubmissionFile.has_valid_extension(filename):
-                log.error(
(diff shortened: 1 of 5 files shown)


ozone/core/api/views.py
```diff
@@ -828,16 +828,20 @@ def handle_pre_create(request):
...
-            if not SubmissionFile.has_valid_extension(filename):
-                log.error(
-                    f'UPLOAD denied for "{token.user}": bad file extension'
-                )
-                return Response(
-                    {'error': 'bad file extension'},
-                    status=status.HTTP_400_BAD_REQUEST
-                )
-
+            """
```
sorinstelian: @danmisailescu : this code does actually work fine and is also covered by functional tests which pass. 

The actual issue you encountered is related to running it in Docker: there is a non-version controlled file (`docker/demo.env`) that should be updated to include the ALLOWED_FILE_EXTENSIONS environment variable (based on the example, source-controlled file `docker/demo.env.example`).

Also, in general it is more useful to log a bug on issues you encounter and include error messages and any other info that can help other developers help you.

PR #539, opened 2019-01-23
Feature/223 export pdf

irina-botez:
Export pdf for article7 and hat

Pdfs can be generated by accessing: **api/submissions/{submission_pk}/export_pdf/**.
This feature requires **reportlab** installed.


sorinstelian:
It looks like the `reportlab` dependency is missing from requirements/base.txt.

Issue #533, opened 2019-01-23
EPIC - Party/Secretariat account page.


nico4:
As a registered user in the System, I can change my account details:

- [ ]  Name
- [ ]  Email
- [ ]  Password
- [ ]  Language

I can see my role in the System and I can contact Secretariat to request a role upgrade (Party_RO -> Party).

Question: Should I see all my colleagues? (E.g. All reporters from Romania).

alexkiro:
> * Password

FWIW, we already have a password change option (from Django), I don't think it's worth adding another one here.

sorinstelian:
@nico4 @melish : taking into account the above comments, we are starting with implementing an endpoint for changing name, email and language.

catalinjitea:
Backend part is done.

We can use `/current-user/<user_pk>` endpoint to update the account details of an user (first_name, last_name, email). For password we will use password change option from Django, like @alexkiro said.

Using the same endpoint, we can view the account of another user. That user should be a secretariat or from the same party. Using `/current-user` we will get account details of the logged in user.

Regarding to `Question: Should I see all my colleagues? (E.g. All reporters from Romania).`, if we decide that we want to see a list of "colleagues" I think we should make another issue.

danmisailescu:
@catalinjitea  Why /current-user instead of returning a JSON with the current user information, returns an array with a single element, and in the code this destructuring hack is needed in order to extract the current user?  This hack assigns to state.currentUser the first element of the array returned from /current-user

setCurrentUser(state, data) {
	[state.currentUser] = data
}
I see we have /api/users/ for returning the list of users.

In the api description, under /api/current-user/ I see "Current User List" . Why? Is this intended to return an array of all the users under one party or managed by a secretariat? In this case what will the convention be? The first element of the array will be the secratariat user who called /current-user? 
![[Untitled].jpg](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5c0fcbafc51ee00011888144/125516c66e82c728ace21e0d46d3a194378b97a58aed8201e81fe086264c22703d46f9692586b32ee6447d070b0a47a8440d5f45ffad8e7ef0e23a25984019af80330b2b9b645bf9fa9015ad00b32a8a4a81beb58949c2ed3c22e18a9d02.jpg)

danmisailescu:
@catalinjitea  I found a problem while testing the endpoint:
PUT http://localhost:8000/api/current-user/2/ 400 (Bad Request)
[...]

PR #546, opened 2019-01-25
Add endpoint for retrieving and updating user details, refs #533

catalinjitea:
Only allow changing name, email, language


alexkiro:
@catalinjitea maybe it would be cleaner to use the existing view `CurrentUserViewSet` ?

catalinjitea:
> @catalinjitea maybe it would be cleaner to use the existing view `CurrentUserViewSet` ?

@alexkiro I thought about it, but that use `CurrentUserSerializer` which contains some extra information that I'm not sure it serves the purpose.

alexkiro:
> @alexkiro I thought about it, but that use `CurrentUserSerializer` which contains some extra information that I'm not sure it serves the purpose.

I think it might be worth just using adding `language` to `CurrentUserSerializer` and setting anything that shouldn't be updated in the `read_only_fields` tuple. Instead of having 3 separate serializers for the user model.

catalinjitea:
> > @alexkiro I thought about it, but that use `CurrentUserSerializer` which contains some extra information that I'm not sure it serves the purpose.
> 
> I think it might be worth just using adding `language` to `CurrentUserSerializer` and setting anything that shouldn't be updated in the `read_only_fields` tuple. Instead of having 3 separate serializers for the user model.

Thanks for the idea. I'll try that.

PR #572, opened 2019-02-05
536 exemptions

catalinjitea:


ozone/core/models/exemption.py
```diff
@@ -56,6 +56,8 @@ class ExemptionApproved(BaseExemption):
 
     decision_approved = models.CharField(max_length=256, blank=True)
 
+    flag_approved = models.BooleanField(default=False)
```
sorinstelian: default should be None (and field type changed accordingly)

ozone/core/models/workflows/accelerated_exemption.py
```diff
@@ -0,0 +1,41 @@
...
+    final_states = ['finalized']
+    editable_data_states = ['data_entry']
+
+    state = AcceleratedExemptionWorkflowStateDescription()
+
+    @xworkflows.transition_check('finalize')
+    def check_finalize(self):
+        return (
+            not self.user.is_read_only and self.user.is_secretariat
+            and self.model_instance.has_approved_exemptions()
```
sorinstelian: actually it could also be that `exemption_approved is False and not self.model_instance.has_approved_exemptions()`

sorinstelian: Actually, disregard the above comment. Since this is for the accelerated workflow, it makes no sense for the OS to create a submission with no approved exemptions. :)

ozone/core/models/workflows/default_exemption.py
```diff
@@ -0,0 +1,57 @@
...
+        return (
+            not self.user.is_read_only and self.user.is_secretariat
+            and self.model_instance.has_nominations()
+        )
+
+    @xworkflows.transition_check('finalize')
+    def check_finalize(self):
+        return (
+            not self.user.is_read_only and self.user.is_secretariat
+            and self.model_instance.has_approved_exemptions()
```
sorinstelian: actually it could also be that `exemption_approved is False and not self.model_instance.has_approved_exemptions()`

sorinstelian: To be more specific: this check should require that the `flag_approved` field is non-null. If it is True, then there should be some exemptions (substance-quantity pairs). It is False, there should be no pairs.

PR #605, opened 2019-02-11
574 Non-Article7 behavior

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -987,26 +998,27 @@ def save(self, force_insert=False, force_update=False, using=None,
...
-            current_submissions = submissions.filter(
-                reporting_period=self.reporting_period
-            )
-            if any([s.data_changes_allowed for s in current_submissions]):
-                raise ValidationError(
-                    _(
-                        "There is already a submission in Data Entry for "
-                        "this party/period/obligation combination."
-                    )
+            if self.is_versionable():
```
sorinstelian: Shouldn't the if clause be *before* select_for_update() ?

ozone/core/models/reporting.py
```diff
@@ -934,19 +950,14 @@ def has_set_approved_flag(self):
...
 
+    def is_versionable(self):
+        return self.obligation.form_type not in self.obligation.NOT_VERSIONABLE
+
     def __str__(self):
         return f'{self.party.name} report on {self.obligation.name} ' \
                f'for {self.reporting_period.name} - version {self.version}'
 
     class Meta:
-        # TODO: this constraint may not be true in the corner case of
```
sorinstelian: Perhaps we should not remove this, but use it differently instead. Let's talk. :)

PR #619, opened 2019-02-13
WIP Allow Secretariat users to enter Date of submission manually

catalinjitea:


ozone/core/models/reporting.py
```diff
@@ -973,6 +974,15 @@ def has_set_approved_flag(self):
     def is_emergency(self):
         return self.flag_emergency
 
+    def can_change_submitted_at(self, user):
```
sorinstelian: also user should not be read-only

ozone/core/serializers.py
```diff
@@ -923,6 +926,11 @@ class Meta:
     def get_reporting_channel(self, obj):
         return getattr(obj.submission.reporting_channel, 'name', '')
 
+    def get_submitted_at(self, obj):
+        submitted_at = getattr(obj.submission, 'submitted_at', None)
+        if submitted_at:
+            return submitted_at.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
```
sorinstelian: Isn't this a date?

catalinjitea: Done.

Issue #620, opened 2019-02-13
can_edit_data & can_upload_files return false positives for party user trying to edit OS submission


sorinstelian:
The flags should return "false" in this case, even if the user has the same party as the one used by OS for creating the submission.

sorinstelian:
Fixed in #621

PR #621, opened 2019-02-13
Fixing behaviour of can_edit_data flag.

sorinstelian:
And implicitly also the can_upload_files flag.


sorinstelian:
Closes #620

Issue #595, opened 2019-02-11
We need a field for type of submission


nico4:
That way we can note that Belgium submitted a letter and not completed A7 forms.  The possible values for “Submission Type” would be as follows:

- A7 Data forms
- Annex to data form 1
- Country Programme data
- Email
- Letter
- Questionnaire

@melish, don't we have "Reporting channel" for this purpose?

mihai-macaneata:
@nico4 reporting channel designates the way the form was submitted and it's not editable by party. Maybe we can modify it ?

sorinstelian:
@nico4 : I would say let's not mix reporting channel and submission type. Isn't `obligation` the field that actually does this?

melish:
Actually, this field models a different information. It does not represent the reporting channel (in this example it would be Email), nor the obligation (Art. 7 in this example), but the contents/format of the alternative submission.
One party could describe their report in a simple text message ("No production/consumption") or attach a document (only the questionnaire, with No for all questions or the Country programme data - a document similar to Art 7 forms, but containing more information required by the Multilateral Fund)

sorinstelian:
The field will be added in the SubmissionInfo tab.

sorinstelian:
Backend part done in https://github.com/eaudeweb/ozone/pull/637. Field name is `submission_format`.

mihai-macaneata:
The field submission_format needs to be in the serializer for api/submissions/pk/submission-info/ 

We also need an endpoint with the avaialable options for the submission format select

catalinjitea:
@mihai-macaneata `submission_format` is already in the submission info serializer.

Also, the endpoint implementation for the available formats will be merged into develop soon.

sorinstelian:
Endpoint available now in develop: `get-submission-formats/`.

@mihai-macaneata : Re-assigning to you. We might also want to look into merging these kinds of general (non-user-specific) endpoints into a single one.

mihai-macaneata:
@melish @sorinstelian @nico4  What about the permissions regarding this field ? is it visible only for secretariat ? In what submission states it can be edited ?

PR #637, opened 2019-02-18
Add field for type of submission

catalinjitea:
refs #595


ozone/core/models/reporting.py
```diff
@@ -1112,6 +1113,15 @@ class SubmissionInfo(ModifyPreventionMixin, models.Model):
     Model for storing submission info.
     """
 
+    @enum.unique
```
sorinstelian: As described in the GitHub issue, we want a model instead of an Enum here.

catalinjitea: Done.

PR #644, opened 2019-02-20
Add endpoint for obligation and period default values

catalinjitea:


ozone/core/models/legal.py
```diff
@@ -56,6 +56,14 @@ def get_current_period(cls):
             end_date__gte=today,
         ).first()
 
+    @classmethod
+    def get_most_recent(cls):
+        return (
+            cls.objects.filter(is_reporting_open=True)
+            .order_by('-end_date')
```
sorinstelian: I'd actually order by start date, to avoid the case of obligations spanning multiple years becoming default.

catalinjitea: Done.

PR #665, opened 2019-02-27
Import of legacy exemptions

catalinjitea:
TODO: Import of legacy exemptions for Exemption obligation.


ozone/core/management/commands/import_exemptions.py
```diff
@@ -0,0 +1,402 @@
...
+        Process the parsed data and insert it into the DB.
+        Only a wrapper, see proccess_submission_exemption_entry and proccess_submission_raf_entry.
+        """
+
+        try:
+            if obligation.pk == 2:
+                return self.proccess_submission_raf_entry(
+                    party_abbr, period_name, rows, obligation, recreate=recreate, purge=purge
+                )
+            # else:
```
sorinstelian: Is this intentional?

catalinjitea: Yes. It will be used for the import of legacy submissions for obligation "Exemption".

ozone/core/management/commands/import_exemptions.py
```diff
@@ -0,0 +1,651 @@
...
+        super().__init__(stdout=None, stderr=None, no_color=False)
+
+        self.periods = {_period.name: _period
+                        for _period in ReportingPeriod.objects.all()}
+        self.parties = {_party.abbr: _party
+                        for _party in Party.objects.all()}
+        self.substances = {_substance.substance_id: _substance
+                           for _substance in Substance.objects.all()}
+
+        self.precision = 10
```
sorinstelian: What is this? :)

ozone/core/management/commands/import_exemptions.py
```diff
@@ -0,0 +1,651 @@
...
+        """
+
+        try:
+            party = self.parties[party_abbr]
+            period = self.periods[period_name]
+        except KeyError as e:
+            logger.critical("Unable to find matching %s: %s", e, rows)
+            return False
+
+        if not period.is_reporting_allowed:
```
sorinstelian: Hmmm, I don't think we should raise this error, we are importing historical submissions here after all.

ozone/core/management/commands/import_exemptions.py
```diff
@@ -0,0 +1,651 @@
...
+            obligation=obligation
+        ).all()
+
+        for sub in qs:
+            logger.info("Deleting submission %s/%s", party.abbr, period.name)
+            for related_data in sub.RELATED_DATA:
+                for instance in getattr(sub, related_data).all():
+                    logger.debug("Deleting related data: %s", instance)
+                    instance.delete()
+            sub.__class__.data_changes_allowed = True
```
sorinstelian: This seems weird. Does it work?

ozone/core/management/commands/import_exemptions.py
catalinjitea: It was supposed to be used for the number of digits after decimal points. I ended up not using it so it should be removed.

ozone/core/management/commands/import_exemptions.py
catalinjitea: It's just a logger that emphasizes the fact that the current period has `
is_reporting_allowed=False`. The problem is that the legacy data contains submissions from 2019.

ozone/core/management/commands/import_exemptions.py
sorinstelian: Yes it does, but it has side effects. As long as `data_changes_allowed` is not properly used anywhere in this script (which really shouldn't happen), it is safe.

Issue #642, opened 2019-02-19
Up to two "data entry" versions should be allowed (one party, one secretariat)


sorinstelian:
* it would be better to allow up to two versions in data entry state (one created by secretariat, one created by party)
* this might imply that cloning a submission in Data Entry should be allowed (if created by a different type of user (party/secretariat) than the other one)

sorinstelian:
This also has a frontend component: the "Another submission already exists in Data Entry stage" message seems to come from the frontend, without making any backend request. This has to be changed.

sorinstelian:
Closed in #654

sorinstelian:
Reopening as it does have a frontend component:
The message "Another submission already exists in Data Entry stage" seems to come from the frontend, without making any backend request. This has to be changed

sorinstelian:
@danmisailescu : should we add a flag (endpoint to be discussed) to signal that the user is allowed to create a new data_entry submission even if another one exists for same party-obligation-period (but was created by a different type of user - i.e. party vs OS)? The current frontend-implemented restriction should be removed or at least changed.

sorinstelian:
Changed priority to "Highest" as this is Article 7-related.

sorinstelian:
I don't believe the current frontend solution takes all cases into account; created https://github.com/eaudeweb/ozone/pull/670 for this.

sorinstelian:
Closed again in #670

PR #670, opened 2019-03-01
Changing check from user id to user type

sorinstelian:
This will also account for situations in which two different users belonging to same Party try to create a new Data Entry submissions


sorinstelian:
Closes #642

PR #674, opened 2019-03-04
Remove duplicated model

catalinjitea:
We use RAFImport for modelling this information


ozone/core/migrations/0093_auto_20190304_1721.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.1.4 on 2019-03-04 14:21
```
sorinstelian: Please re-create this migration

Issue #649, opened 2019-02-21
Exemption - OS-specific buttons "Submit Nomination" and "Finalize" not available


sorinstelian:

sorinstelian:
In short, the workflow for Exemption submissions is different than the one for Article 7 submissions.

Party users can only upload files and view the "Nomination" and "Approved" forms.

Secretariat users can also upload files, but they can also fill in the "Nomination" and "Approved" forms. Both the "Nomination" and "Approved" form can be filled directly with no other prerequisites (i.e. no need to have a file first uploaded to be able to fill them).

After OS has finished filling in the Nomination form, they can submit it by pressing a "Submit Nomination" button. This will transition the Submission to the "Nomination Filled" state, which is not final. Once in this state, they can either fill in the Nomination Approved form, to let the Party know what substances have actually been approved for use, or they can mark the nomination as rejected (described in next paragraph) if the Party obtained no approval.

After OS has finished filling in the Approved form, they can mark it as approved by pressing a "Finalize" button. This will trigger a dialog in which the OS user will be asked whether the nomination is approved or rejected (see end of previous paragraph) - which will be set using the `flag_approved`available in the `submission/<pk>` endpoint.

sorinstelian:
Adding Backend as a component as this does need some changes in the backend too.

sorinstelian:
Added necessary changes in the backend through https://github.com/eaudeweb/ozone/pull/675. 

In order for the `fill_nomination` transition to appear as available, OS-type user should enter at least a substance-quantity pair in the Nomination tab.

After that, the `finalize` transition will appear as available, but will only work if either:
- there is no entry in the "Approved" tab and the submission is marked as not approved
- there are some saved entries in the "Approved" tab and the submission is marked as approved

PR #675, opened 2019-03-05
Making exemption finalize available

sorinstelian:


sorinstelian:
Relates to #649

sorinstelian:
Putting it in done, as it is only a PR.

Issue #612, opened 2019-02-12
Remove unneeded remarks_os/secretariat fields from Submission


sorinstelian:
Also check that the serializers mixins that deal with remarks work as expected.

**LATER EDIT**: they do. :)

sorinstelian:
Actually, these general remarks are used in historical submissions. For such cases, the old remarks data should probably be moved to the questionnaire remarks, but for now it is better to defer it until the OS makes a decision.

melish:
Confirmed: we only need remarks under questionnaire.
The data import should be updated to use those instead of submission.remarks.
Then submission.remarks can be removed.

PR #698, opened 2019-03-07
Legacy data mapping for submitted_at and remarks_os

catalinjitea:


sorinstelian:
@catalinjitea : please see https://github.com/eaudeweb/ozone/issues/612#issuecomment-470628612

PR #716, opened 2019-03-11
Refs. #633. Use a env variable for the TUSD API.

### LINKED ISSUE #633: Subdomain "tusd." should not be hardcoded in source code

alexkiro:
Closes #633


frontend/.env.example
```diff
@@ -1,2 +1,4 @@
+VUE_APP_TUSD_HOST=
```
sorinstelian: @alexkiro : just of out curiosity: was the VUE_APP prefix the only thing needed to make these available in process.env?

frontend/src/components/common/services/api.js
```diff
@@ -9,16 +9,7 @@ const logRequests = process.env.NODE_ENV === 'development'
...
-const windowLocation = window.location.origin.split('//')
-windowLocation[1] = `tusd.${windowLocation[1]}/files/`
-let filesURL = windowLocation.join('//')
-
-/* const TUSD_HOST = 'localhost'
-const TUSD_PORT = 1080
-const _tusd_host = process.env.TUSD_HOST || TUSD_HOST
-const _tusd_port = (process.env.TUSD_PORT && Number(process.env.TUSD_PORT)) || TUSD_PORT
-const filesURL = `http://${_tusd_host}:${_tusd_port}/files/` */
+let filesURL = `${window.location.protocol}//${process.env.VUE_APP_TUSD_HOST}:${process.env.VUE_APP_TUSD_PORT}/files/`
```
sorinstelian: there seems to be a `:` missing after ${window.location.protocol}

frontend/.env.example
alexkiro: Yep, variables that start with [VUE_APP](https://cli.vuejs.org/guide/mode-and-env.html#using-env-variables-in-client-side-code) are the only ones available in the application after build. IIUC the other ones can only be used at build time.

frontend/src/components/common/services/api.js
alexkiro: The wonderful world of JS :laughing: the `protocol` property already includes a trailing `:` for some reason

sorinstelian: Nice :D. Also, I know it's outside the scope of the issue you are solving in the PR, but I think we should also extend these changes to the `apiURL` and `apiBase` variables. What do you think?

alexkiro: > What do you think?

Sounds like a great idea to me :+1: pushed in d08d206

PR #806, opened 2019-04-09
Calculate and generate fixtures for limits

catalinjitea:


ozone/core/management/commands/mkfixtures_limits.py
```diff
@@ -0,0 +1,109 @@
...
+    PartyHistory
+)
+
+
+class Command(BaseCommand):
+    help = "Calculate Limits and generate fixtures."
+
+    OUTPUT_DIR = settings.FIXTURE_DIRS[0]
+
+    def handle(self, *args, **options):
```
sorinstelian: Maybe it should exit with an error if no control measures exist.

ozone/core/management/commands/mkfixtures_limits.py
```diff
@@ -0,0 +1,109 @@
...
+                        ).first()
+                        if not baseline or baseline.baseline is None:
+                            continue
+                        limit = baseline.baseline * cm.allowed
+                        data.append(
+                            self.get_entry(idx, party, period, group, limit_type.value, limit)
+                        )
+                        idx += 1
+                    elif length == 2:
+                        cm1 = cm_queryset_by_limit_type[0]
```
sorinstelian: I think the queryset should be ordered for this to work properly.

ozone/core/management/commands/mkfixtures_limits.py
catalinjitea: Done.

ozone/core/management/commands/mkfixtures_limits.py
catalinjitea: Done, but fixture file hasn't changed.

melish:
Why different rounding rule for BDN_NA5 and NA5_Prod baselines?

image

PR #830, opened 2019-04-16
Zero quantities in Non-Party trade

### LINKED ISSUE #797: Zero quantities in Non-Party trade
There are rows in Non-party trade legacy table without any quantity (in all fields).
For these cases, it is important to keep the remark -> so save it into the overall "Secretariat comments" field in NonParty trade form.
In case of multiple rows for the same party/period and different substances, join the remarks and add the substance name before each remark before merging them.
In case of multiple rows for the same party/period/substance, but different countries, check distinct remarks before merging them.

Examples: for ECE/2007 there are 11 rows, but only 6 distinct remarks. The "Comments secretariat" should contain after import:
```
Carbon Tetrachloride: Export of 0.086 MT to Taiwan not entered since it is part of China 
HCFC-123**: Export of 0.00193 MT to Taiwan not entered since it is part of China
HCFC-124**: Exports of 0.00136MT to Taiwan, 0.0969 MT to Isle of Man not entered since they are part of China and UK respectively.
HCFC-141B: Export of 1.8 MT to Taiwan not entered since it is part of China
Methyl Bromide: Export of 15.482 MT to Taiwan not entered since it is part of China
HCFC-22: Exports of 4.167-Bermuda (UK), 9.145-Cayman Islands (UK), 45.819-Hong Kong (China), 10.465-Isle of Man (UK), 0.22936 MT to Taiwan (China)  not entered since their sovereignty falls under governments t
```

catalinjitea:
closes #797

### DIFF
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -738,7 +738,7 @@ class Article7NonPartyTrade(
         return Party.objects.exclude(id__in=signing_party_ids)
 
     def clean(self):
-        if getattr(self.submission, 'schema_version') != 'legacy' and not (
+        if self.submission.schema_version != 'legacy' and not (
             self.quantity_import_new
             or self.quantity_import_recovered
             or self.quantity_export_new
@@ -755,7 +755,7 @@ class Article7NonPartyTrade(
         # component substance particularly.
         if (
             not self.blend and self.substance and self.trade_party
-            and getattr(self.submission, 'schema_version') != 'legacy'
+            and self.submission.schema_version != 'legacy'
         ):
             non_parties = self.get_non_parties(self.substance.group_id)
             if self.trade_party not in non_parties:
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -738,7 +738,7 @@ class Article7NonPartyTrade(
         return Party.objects.exclude(id__in=signing_party_ids)
 
     def clean(self):
-        if not (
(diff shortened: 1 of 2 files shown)


ozone/core/models/data.py
```diff
@@ -738,7 +738,7 @@ def get_non_parties(group_pk, reporting_period_pk=None):
         return Party.objects.exclude(id__in=signing_party_ids)
 
     def clean(self):
-        if not (
+        if getattr(self.submission, 'schema_version') != 'legacy' and not (
```
sorinstelian: Why not `self.submission.schema_version`?

ozone/core/models/data.py
```diff
@@ -753,7 +753,10 @@ def clean(self):
             )
         # If it's a blend we skip the validation because we will check every
         # component substance particularly.
-        if not self.blend and self.substance and self.trade_party:
+        if (
+            not self.blend and self.substance and self.trade_party
+            and getattr(self.submission, 'schema_version') != 'legacy'
```
sorinstelian: Same here.

ozone/core/models/data.py
catalinjitea: Good point.

catalinjitea: Done.

ozone/core/models/data.py
catalinjitea: Done.

PR #877, opened 2019-04-19
remade table header and changed call to api/aggregations

rexalex:


sorinstelian:
Not sure if this happens already because of the existing code, but just a reminder that in the case of party users, the endpoint should be called by default with the user's party as a parameter (but it should also be possible to view results from other parties if the user wishes so).

PR #901, opened 2019-04-23
Import of LabUse and EssenUse data

### LINKED ISSUE #792: Import of LabUse and EssenUse data
legacy_submissions.xlsx contains rows which sometimes duplicate the LabUse quantity (this applies to Import and Produce tables).
The import of legacy data should calculate quantity_essential_uses by subtracting quantity_laboratory_analytical_uses first. 

Then, for imports, if the substance is Methyl Bromide (id 194), the destination field should be quantity_critical_uses instead of quantity_essential_uses.

There are some cases related to rounding rules for ImpLabUse vs ImpEssenUse.
For instance: CN/2010/100, where ImpEssenUse > ImpLabUse by a very small number. If that happens, use the value in ImpEssenUse, which seem to reflect the total value in ImpNew.

If ImpLabUse is greater than ImpEssenUse (for instance CZ/1998/101), show a visible warning in the import log.

If ImpEssenUse - ImpLabUse > 0.0001, duplicate the remark in both fields (decision_laboratory_analytical_uses and decision_essential/critical).

If -0.0001 <= ImpEssenUse - ImpLabUse  <= 0.0001, save the remark only in decision_laboratory_analytical_uses.

Similar rules for Export (quantity_critical_uses for MeBr) and Production (quantity_lab_use vs quantity_essen_use and quantity_critical_use for MeBr)

Related to #313, #318, #319

catalinjitea:
closes #792.


melish:
LGTM

Issue #921, opened 2019-04-25
Secretariat should be able to directly Finalize an Exemption from initial_state


sorinstelian:
By setting the "emergency" flag.

catalinjitea:
For this we can use the accelerated workflow.

Can we instantiate all submissions created by secretariat with accelerated workflow?

sorinstelian:
I think we can, but we should also talk to @mihai-macaneata to see what the frontend implications are.

sorinstelian:
Decided that OS should first move the "Data Entry" Submission to "Submitted" (optionally adding an attachment), then (if applicable) enter the Nomination and/or Approved data, then move the submisison to "Finalized".

PR #929, opened 2019-05-07
Instantiate the submission's workflow based on user type

catalinjitea:
refs #921


sorinstelian:
Discarding PR as the different workflows approach is not feasible at this point.

Issue #856, opened 2019-04-17
Correlate "Has reported" with treaty ratification data


melish:
Warning about not having reported annex F needs to be tied in with (a) Party having ratified Kigali Amendment – and (b) Party being required to report Annex F – which starts from the year that is 3 months after date of ratification – in any case not earlier than 2019 data.

The same rule applies to the default values for "Annex groups reported in full" checkboxes.

melish:
You can now use /api/parties/246/controlled_groups/?period=2019 to see which groups are mandatory to report on for a given party and period.

melish:
The submission create method (in the backend) should automatically populate has_reported_xx based on the ratification status for each annex group. (@sorinstelian )

melish:
Backend part (See my last comment) still to be done

melish:
closed by #960

melish:
see related #965, #966

PR #960, opened 2019-05-14
Prefill Art7 has reported flags based on ratification status, refs #856

melish:


ozone/core/models/reporting.py
```diff
@@ -1481,7 +1483,7 @@ def save(self, force_insert=False, force_update=False, using=None,
             # (e.g. fast-tracked secretariat submissions).
             # For now we will naively instantiate all submissions with
             # the default article 7 workflow.
-            if self.obligation.form_type == 'exemption':
+            if self.obligation.form_type == FormTypes.EXEMPTION.value:
```
sorinstelian: Nice! :)

PR #1001, opened 2019-05-23
Export PDF for Production and consumption - comparison 2 years

catalinjitea:


ozone/core/api/export_pdf/art7/others.py
```diff
@@ -44,13 +48,14 @@ def get_header(data):
...
 def get_subheader(data):
-    description = _("""Production and Consumption for {period}
-    - Comparison with Base Year""".format(
-        period=data['period']))
-    return Paragraph(
-        description,
-        style=h2_style
+    description = _(
+        "Production and Consumption for {period1} - "
+        "Comparison with {period2} Year".format(
```
melish: Nu mai bine "Comparison with {period2}".format..?

ozone/core/api/export_pdf/art7/others.py
```diff
@@ -210,108 +227,163 @@ def get_prodcons_data(period, parties):
...
+    else:
+        if group in to_report_groups:
+            actual_value = 'N.R.'
+        else:
+            actual_value = '-'
+    return actual_value
+
+
+def get_limit(party, period, group, limit_type, actual_value, to_report_groups):
+    if not isinstance(actual_value, str) and group not in to_report_groups:
```
melish: What if `isinstance(actual_value, str) and group not in to_report_groups`?
This means that party has optionally reported, but a limit is not enforced for that period.

PR #1019, opened 2019-06-05
 Add decision's validity models

catalinjitea:


ozone/core/models/process_agent.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+    contain_technology = models.CharField(max_length=9999)
+
+    class Meta:
+        verbose_name_plural = 'process agent contain technologies'
+        db_table = 'pa_contain_technology'
+
+
+class ProcessAgentUsesValidity(Decision):
```
sorinstelian: I think it's weird (from a user's perspective) to reference a decision like this.

We should have a foreign key to a decision here, I think.

ozone/core/models/process_agent.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class ProcessAgentApplication(models.Model):
+    """
+    Applications of controlled substances as process agents, as approved
+    in table A of decision X/14 and updated periodically by the Meeting of the
+    Parties.
+    """
+
+    decision = models.ForeignKey(
```
sorinstelian: This also seems weird. Foreign key?

ozone/core/models/process_agent.py
catalinjitea: Foreign key to decision seems like a better idea.

ozone/core/models/process_agent.py
catalinjitea: Here the foreign key is to ProcessAgentUsesValidity. Do you mean the field name?

PR #1099, opened 2019-06-24
Whitenoise

mgax:
Use [whitenoise](https://github.com/evansd/whitenoise) to serve static files. It works regardless of debug settings.


mgax:
Don't merge yet because it doesn't actually work for `/reporting`.

mgax:
> Don't merge yet

Ok, not trying to replace the `/reporting/*` route, just adding a flag so that it's served even in debug mode. So now Django can serve all static assets with or without debug.

Issue #1012, opened 2019-05-29
Look and feel for the password reset pages


melish:
https://ozone.edw.ro/admin/password_change/
https://ozone.edw.ro/admin/password_reset/
https://ozone.edw.ro/admin/password_reset/done/

Also the password reset confirmation screen: 
![image](https://user-images.githubusercontent.com/1218168/58558623-ffe55380-8229-11e9-8c09-e201edff2be5.png)

melish:
Buttons don't look right:

![image](https://user-images.githubusercontent.com/1218168/59284183-7c3d5500-8c74-11e9-8b48-faf88bce8241.png)

PR #1100, opened 2019-06-24
Don't track static files

mgax:
Remove existing static files, keep a minimal `.gitignore` in the `/static` folder so that it exists.


melish:
images/background.jpg is definitely needed (for the login page), related to #1012

Not sure about the rest, let's see what @sorinstelian has to say

mgax:
> Or maybe it's ozone/static/images that matters?

It gets copied from `ozone/static/images/background.jpg` to `static/images/background.jpg` when running `collectstatic`.

sorinstelian:
:+1

Issue #1104, opened 2019-06-25
Last modified datetime in dashboard - RAF, EUN


melish:
Does not match the legacy DateUpdate, but is rather the time of the import.

Same for Created by vs DateCreate and submitted_at vs SubmitDate.

related to #1031 

![image](https://user-images.githubusercontent.com/1218168/60075774-bb68ad00-972e-11e9-898a-26c119b7778d.png)

PR #1115, opened 2019-06-26
Be more aggressive in finding create/update dates

mgax:
Some submissions have no records in the `EssenNorm` excel sheet. This patch falls back to the `EssenExemp` sheet.

Addresses https://github.com/eaudeweb/ozone/issues/1104, maybe even fixes it.


sorinstelian:
Looks good, will check whether it has to be applied to other imports as well.

Issue #1091, opened 2019-06-21
E-mail notifications on change submission state


melish:
for main events like Submit, Revise, Recall, Process, Finish, etc.

E-mail should be sent to the secretariat users (define a group in Django or use the is_secretariat flag?) and to the reporter (ORS user + e-mail address from submission info).
Exclude users @example.com

sorinstelian:
Fixed in #1131

PR #1123, opened 2019-06-26
Send emails on workflow changes

mgax:
Addresses https://github.com/eaudeweb/ozone/issues/1091


sorinstelian:
Looks good, merged, will also extend it for the other workflows as well.

Issue #1160, opened 2019-07-25
Data import for MeBrAgreedCriticalUseCategories


melish:
into model ApprovedCriticalUse

PR #1173, opened 2019-07-31
Updated import of exemptions to handle agreed critical use categories, refs #1160

melish:


ozone/core/management/commands/import_exemptions.py
```diff
@@ -677,7 +681,59 @@ def get_exemptions_approved(self, party, period, rows):
...
                 "remarks_os": row['Remark'] if row['Remark'] else "",
-                "is_emergency": row["IsEmergency"] if row["IsEmergency"] else False
+                "is_emergency": True if row["IsEmergency"] else False
             })
 
         return approved_exemptions
+
+    def get_critical_uses_approved(self, party, period, rows):
+        """
+        Parses the MeBrAgreedCriticalUseCategories data for the submission identified by
```
sorinstelian: party & period params aren't actually taken into account :), but the functionality is OK.

PR #1204, opened 2019-08-14
Vagrant provision and test

mgax:
Vagrantfile, provisioning script, and testing script. Tested with the [vagrant-vmck](https://github.com/vmck/vagrant-vmck/) provider but should work with others too.


utility/vagrant_provision.sh
```diff
@@ -0,0 +1,55 @@
...
+  sudo apt-get update -q
+  sudo apt-get install -qy postgresql-9.4
+
+  sudo sed -Ei 's/^(host\s.*)md5/\1trust/' /etc/postgresql/9.4/main/pg_hba.conf
+  sudo systemctl restart postgresql
+)
+
+
+## tusd
+dpkg --list | grep -q postgresql-9.4 && echo "tusd already downloaded" || (
```
sorinstelian: the param to grep seems to be incorrect (should be tusd instead of postgresql-9.4)

mgax: Yep, forgot to fix that conditional, thanks!

mgax:
danke!

PR #1208, opened 2019-08-20
Improvements to the vagrant setup

mgax:
* Build the front-end
* Less verbose output
* A separate env file

The only missing step is importing fixtures, because it takes a reaaaally long time.

Please squash if you merge, there's a lot of little commits that are not worth keeping.


utility/vagrant_provision.sh
```diff
@@ -2,55 +2,76 @@
...
 (
-  echo "Installing dependencies"
+  echo "Installing Python dependencies"
   set -x
 
-  sudo pip3 install -r requirements/tests.txt
+  sudo pip3 install -q -r requirements/tests.txt
+
+  python3 manage.py migrate
+  # python3 manage.py load_initial_fixtures
```
sorinstelian: you should uncomment this line for optimal results.

mgax: Danke, done!

PR #1214, opened 2019-08-21
Adding link for returning to prev page in confirmation page

### LINKED ISSUE #1175: Return the the previous screen after password change
Or, at least, to the dashboard

sorinstelian:
Could not find a more elegant solution (as it actually involves going back two pages in history), suggestions welcome.

Closes #1175


mgax:
Why not just take them to the homepage? I doubt people will be annoyed if we forget their previous location, how often do they change passwords?

melish:
+1 for redirecting to the home page after changing the password

PR #1215, opened 2019-08-21
Export submissions

mgax:
Dump submissions from database to excel file. Fixes https://github.com/eaudeweb/ozone/issues/1201

* [x] Scaffolding to export an excel file
* [x] Filter by party and reporting period; limit rows
* [x] Unit tests
* [x] Cover all columns for "Import" sheet
* [x] And the other sheets
* [x] Usable from Django admin (fixes https://github.com/eaudeweb/ozone/issues/1258)

PR #1251, opened 2019-09-03
Spreadsheet library to load/dump xlsx and directory-with-csv

mgax:
I've made a utility to load and dump spreadsheets in xlsx and csv format. This makes it easier to write tests and manually prepare test data for the export functionality by abstracting the loading and saving of data. It correctly handles numeric values (as float, for now, but this will probably change), dates, and boolean flags. It uses a simple custom format for csv, named internally "csvdir": a directory with a metadata json file that lists the sheets in order, and a csv file for each sheet.

There are no unit tests for the utility, but I've tested it by hand on one of our test xlsx files, making a round-trip from xlsx to csvdir and back, and it works beautifully. It will get test coverage when it's used in test cases of the app, but if you want, I can write a few unit tests for it.

Please have a look, I'd like to use it for https://github.com/eaudeweb/ozone/pull/1215, but it could use its own round of review.


mgax:
Closing in favour of https://github.com/eaudeweb/ozone/pull/1215 which includes this branch

PR #1254, opened 2019-09-10
Allow E2E tests to fail

mgax:
If we don't care so much about the E2E tests, let's configure travis to allow them to fail.


sorinstelian:
My 2 cents: I don't think we should allow them to (more or less) silently fail. There has indeed been a big number of false positives and not enough resources to fix them quickly, but sometimes they can be a good warning for problems (duh!). Let's instead exercise good judgement on what needs quick fixing and what doesn't.

sorinstelian:
Taking my previous comment into account, one possible solution that would allow us to not merge code while having failing tests is to temporarily disable failing tests that do not seem to have a connection with the PR and instead create issues for fixing and re-enabling them, as done in:
https://github.com/eaudeweb/ozone/pull/1257

I have tentatively created a new label for such issues: https://github.com/eaudeweb/ozone/issues?q=is%3Aissue+is%3Aopen+label%3A%22Tests%3A+Skipped+E2E%22 

I believe this approach lets us exercise good reason while also avoiding "bad" practices such as untracked issues and merging code with failing tests. What do you guys think?

mgax:
I think that's a healthy plan. In short, if a test fails and we think it's a false positive, we skip it, and create an issue to fix it later. And we make a point of tracking and fixing these issues.

mihai-macaneata:
Since the tests are fixed and we probably got rid of the most false positive by setting a fixed version for vue-bootstrap, can we close this without merging it ?

mgax:
> tests are fixed

https://github.com/eaudeweb/ozone/pull/1262 ❤️

Issue #1184, opened 2019-08-08
Accessing files uploaded via the backend


melish:
In #1151 there are some new models with a file field (e.g. country profile "other" data).
MEDIA_ROOT is currently set to the tusd repository, but files managed by tusd are not directly accessible.

The question is how to make these files available via the API

sorinstelian:
Closed in #1255

PR #1255, opened 2019-09-10
Changing media root to something non-tusd

sorinstelian:
This will allow public uploads (with public URLs) to be separated from tusd uploads (which are uploaded through tusd, served through API views and permission-controlled).

Relates to #1184

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,6 +85,8 @@ volumes:
     driver: local
   tusd-uploads:
     driver: local
+  media:
+    driver: local
   weblate-data:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -49,6 +49,7 @@ services:
       - ./docker/app.env
     volumes:
      - tusd-uploads:/var/local/tusd_uploads
+     - media:/var/local/media
```
mgax: I think you should add `media` to the list of volumes at the bottom of the file.

PR #1260, opened 2019-09-11
Quantize blend values to `DECIMAL_FIELD_DIGITS`

mgax:
Blend values are calculated and they end up with too many decimal places, and the database refuses to save them. This patch applies quantize again on them.

### DIFF
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -16,7 +16,8 @@ from .utils import (
     model_to_dict,
     decimal_zero_if_none,
     DECIMAL_FIELD_DIGITS,
-    DECIMAL_FIELD_DECIMALS
+    DECIMAL_FIELD_DECIMALS,
+    quantize,
 )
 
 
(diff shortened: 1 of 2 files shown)


ozone/tests/examples/.~lock.art7_submissions.xlsx#
```diff
@@ -0,0 +1 @@
+,mgax,clￄﾃtita,11.09.2019 13:04,file:///Users/mgax/Library/Application%20Support/LibreOffice/4;
```
sorinstelian: Do we want this in the repo?

ozone/core/models/data.py
```diff
@@ -156,6 +157,8 @@ def save(self, *args, **kwargs):
                 attributes['substance_id'] = component.substance.pk
                 attributes['blend_item_id'] = self.pk
                 attributes.update(field_dictionary)
+                attributes['quantity_feedstock'] = quantize(attributes['quantity_feedstock'])
```
sorinstelian: Are these lines needed?

ozone/tests/examples/.~lock.art7_submissions.xlsx#
mgax: Whoopsie :)

Issue #1226, opened 2019-08-23
Allow saving invalid records during data entry state


melish:
This implies that we have to do some extra checks when the forms are submitted (don't assume that all records are already valid).

sorinstelian:
Backend part done in #1279. Backend checks are now performed at the submit transition (rather than at data save). The transition will fail with a relevant error message if the conditions in #81 are not satisfied.

PR #1279, opened 2019-09-18
Perform import/export validity check at submit time

sorinstelian:
This allows temporarily incomplete data to be saved during data entry.

Relates to #1226

### DIFF
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -610,18 +610,13 @@ class BaseImportExportReport(models.Model):
 
             # And finally verify that, for each substance,
             # sum of totals > sum of quantities
-            valid = all(
-                [
-                    sums['totals_sum'] >= sums['quantities_sum']
-                    for sums in sums_dictionary.values()
-                ]
-            )
-            if not valid:
-                raise ValidationError(
-                    f'For each substance that has no {party_field}, '
-                    f'the sum of quantities across all data entries should be '
-                    f'less than the sum of totals'
-                )
+            for id, sums in sums_dictionary.items():
+                if sums['totals_sum'] < sums['quantities_sum']:
+                    raise ValidationError(
+                        f'For {Substance.objects.filter(id=id).first()}, the '
+                        f'sum of quantities across all data entries is greater '
+                        f'than the sum of totals.'
+                    )
 
     class Meta:
         abstract = True
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -545,21 +545,20 @@ class BaseImportExportReport(models.Model):
         # one entry with a src/dst country.
         partyless_substances = set()
         partyful_substances = set()
+        # Only taking into account substance entries (as there already are
+        # substance entries auto-generated from blends - and calculating them
+        # again is redundant)
+        substance_entries = cls.objects.filter(submission=submission).exclude(
+            blend__isnull=False
+        )
 
-        for entry in cls.objects.filter(submission=submission):
+        for entry in substance_entries:
             substance = getattr(entry, 'substance', None)
-            blend = getattr(entry, 'blend', None)
-            if getattr(entry, party_field, None) is None:
-                if substance:
-                    partyless_substances.add(substance.id)
-                elif blend:
-                    partyless_substances.update(blend.get_substance_ids())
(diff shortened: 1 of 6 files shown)


ozone/tests/test_art7_import.py
```diff
@@ -197,6 +197,39 @@ def test_create_multiple_source_party_duplicate(self):
...
+        self.assertEqual(result.status_code, 201, result.json())
+
+        result = self.client.post(
+            reverse(
+                "core:submission-call-transition",
+                kwargs={"pk": submission.pk},
+            ),
+            {"transition": "submit", "party": submission.party.pk},
+        )
+        self.assertEqual(result.status_code, 422, result.json())
```
mgax: I gather that you're asserting that the backend returns an error. Why do we expect an error here? Would it make sense to fix the error, submit again, and assert for success?

ozone/tests/test_art7_import.py
```diff
@@ -197,6 +197,39 @@ def test_create_multiple_source_party_duplicate(self):
         )
         self.assertEqual(result.status_code, 422, result.json())
 
+    def test_create_inconsistent_data(self):
+        # Test according to https://github.com/eaudeweb/ozone/issues/81
```
mgax: It would be nice to provide a short description of what the test checks so I don't have to copy-paste a link and go to GitHub. But keep the link, it's useful :)

ozone/tests/test_art7_import.py
```diff
@@ -197,6 +197,39 @@ def test_create_multiple_source_party_duplicate(self):
...
+        submission = self.create_submission()
+
+        # This has 'quantity_total_new': 40.0
```
mgax: You can write this as an assertion, that way I don't have to trust that nobody messed with `ART7_IMPORT_DATA` in the meantime :)

ozone/core/models/workflows/default.py
```diff
@@ -129,9 +129,9 @@ def before_submit(self, *args, **kwargs):
         """
         Called right before the transition is actually performed.
         """
-        # Also validate imports and exports data (will raise a validation error
+        # Validate imports and exports data (will raise a validation error
         # if data is not consistent).
-        return True
+        self.model_instance.check_imports_exports()
```
mgax: So this used to return `True`, now it returns `None`, i suppose that's ok?

ozone/core/models/data.py
```diff
@@ -518,6 +522,107 @@ class BaseImportExportReport(models.Model):
         max_length=512, blank=True
     )
 
+    @classmethod
+    def validate_import_export_data(cls, submission):
```
mgax: I would suggest extracting a couple of helper functions from this one, e.g. one to encapsulate the first big loop and calculate partyful and partyless substances. Give them relevant names and the whole thing becomes easier to mentally parse.

ozone/core/models/data.py
```diff
@@ -518,6 +522,107 @@ class BaseImportExportReport(models.Model):
...
+
+            # And finally verify that, for each substance,
+            # sum of totals > sum of quantities
+            valid = all(
+                [
+                    sums['totals_sum'] >= sums['quantities_sum']
+                    for sums in sums_dictionary.values()
+                ]
+            )
+            if not valid:
```
mgax: You can write the `all` as a loop and throw an error at the first invalid value. That way you can provide more info to the user.

ozone/core/models/reporting.py
```diff
@@ -821,6 +821,19 @@ def is_submittable(self):
...
 
+    def check_imports_exports(self):
+        """
+        Performs checks on all imports and exports data in this submission
+        based on the validation rules described in
+        https://github.com/eaudeweb/ozone/issues/81/
+        If validation fails, this will raise a validation error.
+        """
+        if self.obligation.obligation_type == ObligationTypes.ART7.value:
+            if hasattr(self, "article7imports") and self.article7imports:
```
mgax: This looks scary. Maybe dispatch based on `isinstance`? Or override the method in subclasses?

ozone/core/models/data.py
sorinstelian: Indeed, done.

PR #1285, opened 2019-09-18
Allow select all parties or periods in reports, fixes #1217

### LINKED ISSUE #1217: Allow users to download reports without selecting a party
In this case, all parties will be considered selected (this can be handled by the backend).

melish:


ozone/core/api/export_pdf/reports/prodcons.py
```diff
@@ -179,17 +179,22 @@ def get_prodcons_flowables(submission, periods, parties):
 
 def _get_table_data(party, period, prodcons_qs, submission, date_reported, all_groups):
     table_data = {}
-    history = PartyHistory.objects.get(
-        party=party,
-        reporting_period=period
-    )
+    try:
```
sorinstelian: I think there should also be a `population = party_type = None` line before the try-except block.

Issue #1292, opened 2019-09-20
Separate upload locations by model


melish:
A few models related to country profiles (Website, LicensingSystem, OtherCountryProfileData) have file fields. It would be good not to store the uploaded files in the same location (all in MEDIA_ROOT), but to create a folder for each model (based on the model name).
Check the country_profile.user_directory_path method

Issue #1291, opened 2019-09-20
File and URL fields on LicensingSystems model should have some titles


melish:
as in `<a href="$url">$title</a>`
To allow entering a good description for the uploaded files or external URLs.
This files/url's will be presented on the main website, in the country profile.

mgax:
So this means creating `title` fields for those models, that the user has to fill in, and that will be exposed via the API to be displayed on the country profile, right?

PR #1298, opened 2019-09-23
File upload

mgax:
Save files to subfolders; require title for licensing system files and urls.

Fixes https://github.com/eaudeweb/ozone/issues/1292; fixes https://github.com/eaudeweb/ozone/issues/1291


ozone/core/migrations/0173_auto_20190806_1635.py
```diff
@@ -15,12 +19,12 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='othercountryprofiledata',
             name='file',
-            field=models.FileField(blank=True, null=True, upload_to=ozone.core.models.country_profile.user_directory_path),
+            field=models.FileField(blank=True, null=True, upload_to=dummy_upload_to),
```
sorinstelian: Why change older migrations? Making a new one does not work?

PR #1299, opened 2019-09-24
Aggregations - fields for calculated qps and laboratory production/consumption

sorinstelian:


ozone/core/models/aggregation.py
```diff
@@ -320,13 +344,22 @@ def get_export_quarantine(self):
             return self.export_quarantine
         return Decimal(0.0)
 
+    def get_export_or_production_quarantine(self):
+        if self.production_quarantine >= self.export_quarantine:
+            return self.export_quarantine
+        elif self.production_quarantine < self.export_quarantine:
+            return self.production_quarantine
+        else:
```
mgax: When would this `else` ever be reached? It seems that the `if` and `elif` conditions cover all the possibilities.

Issue #1236, opened 2019-08-23
Email notifications on change submission state


melish:
related to #1091 

- [x] Change the text to:
"$Party_name has SUBMITTED its report on Article 7 - Data Reporting for 2018 - this being submission number # 2. Submitted by test_mv."
- [x] Send a single message, using To: secretariat users and Cc: party users

PR #1300, opened 2019-09-24
Workflow notification emails

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1236

* Send a single message with "to" secretariat, "cc" party users and the email in the submission. (Sending to all party users is new, this is what we want, no?)
* Helper function to dig out the `site_name`, hopefully we'll only ever have one site per database.

### DIFF
--- a/ozone/core/models/workflows/emails.py
+++ b/ozone/core/models/workflows/emails.py
@@ -5,8 +5,8 @@ from ozone.core.utils.site import get_site_name
 
 
 def allow_sending_to(email):
-    # if re.search(r'@example\.(com|org)$', email):
-    #     return False
+    if re.search(r'@example\.(com|org)$', email):
+        return False
 
     return True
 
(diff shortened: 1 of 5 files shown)


ozone/core/models/workflows/emails.py
```diff
@@ -1,31 +1,37 @@
...
 from ozone.core.models import User
 from ozone.core.email import send_mail_from_template
+from ozone.core.utils.site import get_site_name
 
 
-def skip(email):
-    if re.search(r'@example\.(com|org)$', email):
-        return True
+def allow_sending_to(email):
+    # if re.search(r'@example\.(com|org)$', email):
```
melish: This one was needed to prevent send errors during the beta testing (when all users have dummy email addresses associated)

mgax: Whoops, I commented it out for testing, forgot to uncomment.

Issue #1293, opened 2019-09-20
Separate uploads by public and private


sorinstelian:
Related to #1292

mgax:
Conclusion: we make `MEDIA_ROOT` private by default and create a `/public` folder inside which is served without authentication as public files. Country profile files will live in this public folder, while submissions will not, and they must be accessed via the submissions API.

mgax:
I'm leaving the issue open for a bit more, to update the nginx/apache config for all deployments.

mgax:
Ok, `ozone-staging.edw.ro`, `ozone.edw.ro` and `ors.ozone.unep.org` are up to date with only serving public media from nginx or apache.

PR #1303, opened 2019-09-24
Public media

mgax:
Addresses https://github.com/eaudeweb/ozone/issues/1293

* ~Move the serving of media files to the nginx from docker-compose~
* Change models to save public files in `MEDIA_ROOT + '/public'`
* Only serve the `media/public` folder, both from nginx and from the django app when `DEBUG=True`

Post-merge:
* ~Make sure the docker-compose nginx does in fact serve the files~
* [ ] Check all servers to make sure their local nginx/apache configs are up to date


docker-compose.yml
```diff
@@ -21,8 +21,8 @@ services:
           server {
             listen 8000;
 
-            location /media/ {
-              alias /opt/ozone/media/;
+            location /media/public/ {
```
alexkiro: Note that this nginx is just doing load balancing for multiple apps here. And simply acts as TCP stream proxy, so I don't believe this will work, since it won't be aware of the HTTP/WSGI request, so `location` will likely not be valid?

melish: Thanks @alexkiro you are probably right. What @mgax was trying to move as much as possible from the host nginx.conf to the source repo (ideally reduce it to a simple proxy + ssl terminator).
Even if we make the nginx from docker-compose listen on http, I'm not sure how this will work for weblate and tusd.

mgax: > Even if we make the nginx from docker-compose listen on http, I'm not sure how this will work for weblate and tusd.

They work over http, don't they? Even if they use websockets nginx is able to proxy that.

mgax:
After reading the nginx configs more carefully, I realise the nginx in docker-compose is only intended to be a TCP proxy. If we're to convert it to a full http proxy, it makes sense to use it for tusd, mailtrap and weblate too, perhaps unifying the `tusd` and `translate` subdomains into the main domain as subfolders. I think that would be a good thing, but it's already outside the scope of this task, so I've amended my PR to simply upate the existing example nginx config.

sorinstelian:
Merging so we can test on staging.

Issue #1077, opened 2019-06-18
Allow pre-defined attachments when sending e-mails


melish:
- raw data report
- production and consumption report
- other custom files (e.g. form templates)

configurable at template level

mgax:
Are you referring to workflow notification emails? Where do I find the reports you're talking about?

PR #1304, opened 2019-09-24
Model and serializer for email attachments

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1077

* [x] Model and serializer for static attachments
* [x] Serializer for generated attachments, so we don't hardcode them in the UI
* [x] Extend `/api/submissions/$id/email/` to accept both kinds of attachments

I have a couple of concerns:
* Should we save the attachments on the [Email model](https://github.com/eaudeweb/ozone/blob/2911e6de24cdc6ec64dc3fb23f5fa50adbb979c9/ozone/core/models/email.py#L15)? Perhaps as a JSON field, containing filename and size? e.g. `attachments = [{"filename": "something.pdf", "size": 1234, "mime-type": "application/pdf"}, {"filename": "prodcons_EU_2009.pdf", "size": 5678, "mime-type": "application/pdf"}]`
* ~Should we unify the attachments in the API? I expect the front-end doesn't care which attachments are generated and which are static.~ Semi-done: `GET /api/email-templates/` returns separate lists, `attachments` for static files linked to the template and `generated_attachments` for the reports; I expect it will be straightforward for the UI to merge the lists. On the other end, when sending an email (`POST /api/submissions/2/email/`), both kinds of attachments go into the `attachments` field, with an extra argument `source` to differentiate between the types.

Here is what the API calls look like:

* fetching email templates - `GET /api/email-templates/`
    ```json
    [
      {
        "id": 1,
        "attachments": [
            {
              "id": 1,
              "filename": "peșcheş.txt",
              "title": "peșcheş!",
              "source": "email_template_attachment"
            }
        ],
        "generated_attachments": [
            {
              "id": "art7_raw",
              "filename": "art7raw_{party}_{period}.pdf",
              "title": "Raw data reported - Article 7",
              "source": "generate_report"
            },
            {
              "id": "prodcons",
              "filename": "prodcons_{party}_{period}.pdf",
              "title": "Production and consumption - comparison with base year",
              "source": "generate_report"
            }
        ],
        "name": "EN_no_issues",
        "subject": "[A7 2018] Country_XXXX: data reporting for 2018",
        "description": "--- description ---"
      }
    ]
    ```
* send an email - `POST /api/submissions/2/email/`
    ```json
    {
      "body": "--- message body ---",
      "cc": [],
      "from_email": "admin@example.com",
      "subject": "[A7 2018] Country_XXXX: Data Reporting",
      "to": ["nobody@example.com"],
      "attachments": [
        {"source": "email_template_attachment", "id": 1},
        {"source": "generate_report", "id": "prodcons"}
      ]
    }
    ```


melish:
> Should we unify the attachments in the API? I expect the front-end doesn't care which attachments are generated and which are static.
Yes, there are no reasons for the UI to care

ozone/core/serializers.py
```diff
@@ -1818,27 +1823,102 @@ class Meta:
...
+    period = submission.reporting_period
+
+    if report == Reports.ART7_RAW.value:
+        art7 = Obligation.objects.get(
+            _obligation_type=ObligationTypes.ART7.value)
+        data = export_pdf.export_submissions(art7, [submission])
+
+    elif report == Reports.PRODCONS.value:
+        data = export_pdf.export_prodcons(
+            submission=submission,
```
melish: submission is mutually exclusive with (periods, parties)
see `get_prodcons_data` in `prodcons.py`:

```
    parties = [submission.party] if submission else parties
    periods = [submission.reporting_period] if submission else periods
```

Just to avoid confusion, use None for periods and parties parameters.

ozone/core/migrations/0205_email_template_attachment.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 2.1.4 on 2019-09-24 14:55
+
+from django.db import migrations, models
+import django.db.models.deletion
+import ozone.core.models.file
+
+
+class Migration(migrations.Migration):
```
melish: Remember about #1345
It would be easier to deploy the "squashed" release when migrations are in sync (prod/staging server vs HEAD)

ozone/core/models/email.py
```diff
@@ -24,21 +25,41 @@ class Email(models.Model):
         null=True, on_delete=models.SET_NULL,
     )
 
-    def send_email(self):
+    def send_email(self, attachments=[]):
```
alexkiro: You have a mutable default argument here.

ozone/core/serializers.py
mgax: Thanks, done.

ozone/core/migrations/0205_email_template_attachment.py
mgax: That will not be a problem because the squashed migrations will coexist with the old ones until all systems are up to date - https://docs.djangoproject.com/en/2.2/topics/migrations/#squashing-migrations

> The recommended process is to squash, keeping the old files, commit and release, wait until all systems are upgraded with the new release (or if you’re a third-party project, just ensure your users upgrade releases in order without skipping any), and then remove the old files, commit and do a second release.

ozone/core/models/email.py
mgax: Yeap, thanks. Changing to `None` since that's what `EmailMultiAlternatives` uses as default too.

mgax:
@melish should we save info about the attachments (filename, size, mime type) on the outgoing email model? It might be nice to display this information in the email history.

melish:
> 
> 
> @melish should we save info about the attachments (filename, size, mime type) on the outgoing email model? It might be nice to display this information in the email history.

Just the attachment type (Art 7 Raw, Prod/Cons, etc.) should be enough for the users.

mgax:
Alrighty, I've saved title, filename and size. Since we already have all the info handy. The UI doesn't have to display all of it.

PR #1307, opened 2019-09-24
Fetch external data on each save.

sorinstelian:
This should avoid some weird corner cases without really affecting performance.


melish:
Lgtm, but why was the change needed?

Issue #1331, opened 2019-09-28
Unsupported operation when exporting the legaxy xlsx


melish:
https://sentry.edw.ro/eaudeweb/ozone/issues/53
related to #1258, #1215

PR #1342, opened 2019-09-30
Use a temp file on disk so that sendfile works

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1331


ozone/core/admin.py
```diff
@@ -451,11 +450,13 @@ def delete_model(self, request, obj):
...
     def export_legacy_xlsx(self, request, queryset):
         data = export_submissions(queryset)
-
-        with capture_temp_file('legacy_submissions.xlsx') as (path, output):
-            data.dump_xlsx(path)
-
-        return django_file_response(output)
+        output = tempfile.NamedTemporaryFile()
+        data.dump_xlsx(output.name)
+        return FileResponse(
```
alexkiro: I don't believe that using a temporary file + `sendfile` provides any benefits here, so I would switch this to a simple HttpResponse with the appropriate headers set, to avoid the extra complexity.

One sendfile operation would be more efficient than the combination of one read operation + one write operation, however since this is being generated in memory, here we are doing one write into a temp file **and** a sendfile. Which in theory should be much less efficient than a simple write directly into the HttpResponse. 

_Disclaimer: I haven't ran any actual benchmarks or tests, though._ :P

mgax: Fair enough, but:
* The obvious openpyxl API works with files on disk, and a quick check indicates it's not able to save a bytestream in memory - maybe I'm wrong though
* The bulk of the work is querying the database and generating the data
* There's a useful performance advantage in dumping the file to disk and freeing Python memory when the view function is done

I think the main argument for this tempfile approach is code simplicity :)

PR #1350, opened 2019-10-01
Nominations and approval - read only for parties, closes #1172

### LINKED ISSUE #1172: Exemption nominations and approval - read only for parties
Countries should not be able to use the data entry webforms for EUE/CUE nominations and EUE/CUE approvals – they should only be able to submit files.  Only Secretariat should be able to enter the numbers.  Currently, when logged in as Kenya, we can enter EUE/CUE nominations and EUE/CUE approvals as Kenya.

razvanMiu:


melish:
Probably the placeholder in each table ("Please use the form on the right sidebar to add substances") should by dynamic, also (either visible only for secretariat users, or replaced with "This table shows substances and amounts once the secretariat processes your submission").

frontend/src/components/exemption/form-components/Add.vue
```diff
@@ -64,7 +64,7 @@ import { getAlerts } from '@/components/common/dataDefinitions/alerts'
 export default {
 
   props: {
-    tabName: String
+    tabName: String,
```
melish: This comma produces a linter error

frontend/src/components/exemption/form-components/DefaultAside.vue
```diff
@@ -6,6 +6,6 @@ export default {
   mixins: [DefaultAside],
   components: {
     add: Add
-  }
+  },
```
melish: This comma produces a linter error

PR #1354, opened 2019-10-01
Fixed flags not saving, closes #1223

### LINKED ISSUE #1223: Provisional flag is not saved when saving a read-only submission
e.g. Open a finalized submission using secretariat, check the Provisional flag, save and refresh.

razvanMiu:
The issue was that the flags tab's status wasn't updated when a flag was ticked.


melish:
Looks good to me, except one thing, which may or not may be related to this PR:

When creating a submission as secretariat, before Submit there is a popup which allows changing the has_blanks flag:

image


image

Check some previously unchecked flags and submit:

image

After submit, the flags seem to be correctly updated:

image

The PUT request seems to have the right values:
image

but after refresh, the flags are reverted to the initial values:

image

melish:
Created #1359, related to the comment above, as it seems to be a backend issue.

PR #1357, opened 2019-10-02
Nginx conf for production server

melish:
where we don't have subdomains


config/nginx/nginx_unep.conf.example
```diff
@@ -0,0 +1,83 @@
...
+      root /var/www/html/letsencrypt;
+  }
+
+  access_log /var/log/nginx/access.log main;
+}
+
+server {
+    server_name ors.ozone.unep.org;
+    listen 443 ssl; # managed by Certbot
+    ssl_certificate /etc/letsencrypt/live/ors.ozone.unep.org/fullchain.pem; # managed by Certbot
```
mgax: Just to mention, nowadays there are alternatives to ngninx+certbot, e.g. [Traefik](https://traefik.io/), which can obtain their own Letsencrypt certificates.

PR #1362, opened 2019-10-02
Aggregating by group in the MT ViewSet.

### LINKED ISSUE #1327: Allow aggregation by group on aggregations MT endpoint

sorinstelian:
Closes #1327

### DIFF
--- a/ozone/core/api/views.py
+++ b/ozone/core/api/views.py
@@ -747,8 +747,11 @@ class AggregationViewSet(viewsets.ReadOnlyModelViewSet):
 
 class AggregationMTViewSet(AggregationViewSet):
     # This view displays Metric Tons data for all annex groups
-    # TODO: prefetch_related?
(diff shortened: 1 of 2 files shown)


ozone/core/api/views.py
```diff
@@ -645,6 +639,88 @@ def get_queryset(self):
...
+        all_values = queryset.values(
+            *fields, 'party', 'reporting_period', self.group_field
+        )
+        values = []
+        for period in periods:
+            # Use a list of dictionaries for in-memory storage to optimize
+            # DB queries.
+            values_list = [
+                value for value in all_values
+                if value['reporting_period'] == period
```
mgax: There's an opportunity to optimize things: make a single pass through `all_values` and append them to buckets based on `reporting_period`. But only if the API call is slow.

sorinstelian: Yes, will do, as it was originally implemented as you suggest and it was faster. Merging like this and will open a separate issue.

PR #1365, opened 2019-10-02
Allow saving invalid records; Validation only on submit, closes #1226

### LINKED ISSUE #1226: Allow saving invalid records during data entry state

razvanMiu:


melish:
A few remarks:

- [ ] The mandatory fields for Submission Info are now optional, as well. They should remain required, at least before Submitting

- [ ] because of that, when any of `Date of submission`, `Name of reporting officer`, 'Email' etc are empty, the Submit button is hidden and users will definitely be confused

- [ ] In case there are validation errors and users press Submit, they should receive a notification immediately (not after pressing OK in the submit modal dialog)

PR #1378, opened 2019-10-07
Submission not closing after pressing the Process button, closes #1340

### LINKED ISSUE #1340: Don't close the submission after pressing the Process button
When pressing `Process`, the submission is closed and the user is taken to the Dashboard (instead of staying on the submission).

Related to #1197

razvanMiu:


melish:
Merging, will test on staging because this doesn't happen in development mode

PR #1383, opened 2019-10-07
No submit without any attachment; Updated polyfills.js in according to ESLint rules; closes #1195

### LINKED ISSUE #1195: Don't allow to submit without any attachment
For any obligation other than Art7, HAT, ~Exemption~ and RAF.
Only applicable to parties (secretariat should still be able to submit without attaching files).
I propose to implement only in frontend, otherwise we'll have trouble when importing legacy data.
You can also disable this check while in development mode.

melish:
Re-created #1381


melish:
It seems that the validation also applies to secretariat user. Please see comment on #1195

> secretariat should still be able to submit without attaching files

PR #1388, opened 2019-10-15
Extract `ReportingPeriod.is_control_period`

mgax:
This is used in 2 places already and I need it for the limits UI too.


ozone/core/management/commands/mkfixtures_limits.py
```diff
@@ -39,12 +39,10 @@ def handle(self, *args, **options):
             party = party_history.party
             party_type = party_history.party_type
             period = party_history.reporting_period
-            if period.name in ["C1999", "C2000", "C2001"]:
+            if period.is_control_period:
```
sorinstelian: The old condition checked here is different than the new one. The new one looks correct though.

Issue #1408, opened 2019-10-22
Validation rules for Art7 imports and exports


melish:
Should not include the recovered amounts.
That is: the sum of individual components should be less than or equal to the *New* Imports/Exports.
The warning message from the validation tab should be updated accordingly.

This also involves changes in the backend validate_import_export_data (data.py)

related to #440, #81, #908, #822 etc

sorinstelian:
Backend part done in #1409 ; backend error message remains the same as it does not reference recovered quantities.

PR #1409, opened 2019-10-23
Removing quantity_total_recovered from validation

sorinstelian:
It should not actually be taken into account.

Refs #1408

### DIFF
--- a/ozone/core/models/data.py
+++ b/ozone/core/models/data.py
@@ -548,7 +548,7 @@ class BaseImportExportReport(models.Model):
         validation criteria (see https://github.com/eaudeweb/ozone/issues/81)
         """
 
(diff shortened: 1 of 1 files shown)


ozone/core/models/data.py
```diff
@@ -548,7 +548,7 @@ def validate_import_export_data(cls, submission):
         validation criteria (see https://github.com/eaudeweb/ozone/issues/81)
         """
 
-        totals_fields = ['quantity_total_new', 'quantity_total_recovered']
+        totals_fields = ['quantity_total_new',]
```
melish: I think you should exclude `quantity_total_recovered` from the quantity_fields (see check a few lines below).

Issue #1396, opened 2019-10-16
Improve speed of calculate_baselines and calculate_limits


melish:
see  ozone/core/calculated/baselines.py and limits.py
It takes forever to recalculate all baselines, probably because of too many SQL queries.
Maybe all data could be pre-loaded and processed from memory.

For instance, when recalculating baselines for a single party and a single annex group, out of 470 db queries there are about 80 queries from baselines, 80 from limits, 150 from party/party_history (including several for parties CN - see is_china_or_brazil or is_european_union - executed 20 times)

sorinstelian:
Fixed in #1423

PR #1423, opened 2019-10-25
1396 baseline optimizations

sorinstelian:
Reducing the number of DB queries and the total computing time for baseline calculations.

Refs #1396

### DIFF
--- a/ozone/core/calculated/baselines.py
+++ b/ozone/core/calculated/baselines.py
@@ -3,7 +3,6 @@ from decimal import Decimal
 from functools import lru_cache
 import logging
 
-from django.db import models
 from django.db import transaction
 from django.template.response import TemplateResponse
 from django.contrib import messages
@@ -78,7 +77,7 @@ class BaselineCalculator:
             for _period in ('1989', '2009', '2010')
         }
 
-    @lru_cache(maxsize=16)
+    @lru_cache(maxsize=2)
     def _get_prodcons_objects(self, group, party):
         prodcons_objects = {p: None for p in self.reporting_periods}
         for p in ProdCons.objects.filter(group=group, party=party):
@@ -91,7 +90,7 @@ class BaselineCalculator:
         )(group, party)
         return func(party, group, periods) if func else None
 
-    @lru_cache(maxsize=128)
+    @lru_cache(maxsize=2)
     def _get_prodcons(self, party, group, period_name):
         prodcons_objects = self._get_prodcons_objects(group, party)
         p = prodcons_objects.get(period_name, None)
@@ -485,7 +484,7 @@ class BaselineCalculator:
             func = self.average_production_bdn
         return func, periods
 
-    @lru_cache(maxsize=16)
+    @lru_cache(maxsize=2)
     def _get_aggregation_from_submission(self, submission_id):
         """
         Helps cache result of expensive call made in _prod_cons_gwp().
--- a/ozone/core/calculated/baselines.py
+++ b/ozone/core/calculated/baselines.py
@@ -32,6 +32,7 @@ class CalculationError(Exception):
 class BaselineCalculator:
 
     def __init__(self):
+        self.current_period = ReportingPeriod.get_current_period()
         self.groups = {
             _group.group_id: _group
             for _group in Group.objects.all()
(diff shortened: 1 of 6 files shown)


ozone/core/calculated/baselines.py
```diff
@@ -54,44 +63,73 @@ def __init__(self):
                 self.reporting_periods['1989']
             )
         ]
+        # Will be set when get_baseline() is called and unset at call completion
```
mgax: This feels really fragile and it will make the code harder to follow. I'd suggest several other approaches:

* Keep the cache out of the `BaselineCalculator` class (e.g. a lru cache as you suggest below)
* Create a namespace! (["Namespaces are one honking great idea -- let's do more of those!"](https://www.python.org/dev/peps/pep-0020/)) e.g. `self.prodcons_cache = defaultdict(dict)`, where the key can be your `prodcons_objects_key`, and then `self.prodcons_cache.pop(key)` at the end, to conserve memory.

ozone/core/calculated/baselines.py
```diff
@@ -54,44 +63,73 @@ def __init__(self):
...
+                if _party not in eu_members_for_period
             ]
         self.new_eu_member_states_since = {
             _period: _new_eu_member_states_since(_period)
             for _period in ('1989', '2009', '2010')
         }
 
+    def _populate_prodcons_cache(self, group, party):
+        # TODO: this could be lru_cached and also the cache key could be just
+        # the party (needs more testing).
```
mgax: > needs more testing

You realise this will stay forever as a TODO, no? :) As a general approach, I think it's better to write the code so that it fails in a clear manner, perhaps throw a `RuntimeError` in the right place, so that it's easy to debug when it does fail.

ozone/core/calculated/baselines.py
```diff
@@ -54,44 +63,73 @@ def __init__(self):
...
+        if (group, party) != self.prodcons_objects_key:
+            prodcons_values = ProdCons.objects.filter(
+                group=group, party=party
+            ).values(
+                'reporting_period__name', 'is_article5', 'is_eu_member',
+                'submissions',
+                *[
+                    f.name for f in ProdCons._meta.get_fields()
+                    if isinstance(f, models.DecimalField)
+                ]
```
mgax: Can you extract this list as a variable? Too much nesting here.

melish:
The improvements seem pretty good for annexes A,B,E (number of queries reduced from 100+ to 18), but not enough for C/I and F, where there are still 1000+ queries made for one party and one annex group.

These are worth looking into:
- baseline types (300+)
- limit_prod_cons (300+)
- 'EU' party (80)
- "data" models for a submission (imports, exports, destroyed, etc. - because of a queryset filtering by individual annex groups) - 230 queries in total

- maybe groups - where group_id in ('AI', 'AII', 'BI', 'BII', 'BIII', 'CI', 'CII') (~10 queries)
- maybe party_history (~10 queries)
- maybe submission (~10 queries for 3 submissions)

ozone/core/calculated/baselines.py
```diff
@@ -54,44 +63,73 @@ def __init__(self):
...
+            self.prodcons_objects_key = (group, party)
+
     def get_baseline(self, baseline_type, group, party):
+        self._populate_prodcons_cache(group, party)
         func, periods = getattr(
             self, 'baseline_' + baseline_type
         )(group, party)
         return func(party, group, periods) if func else None
 
     @lru_cache(maxsize=128)
```
mgax: I know this is pre-existing code, but since you have the problem in your head already, could you write a short comment, describing how big the cache should be? e.g. `we estimate len(parties) * len(groups) ~= 100 cache entries for a calculation of xyz baselines`. If it makes sense. If not, forget what I said.

ozone/core/calculated/baselines.py
sorinstelian: It did not remain as a TODO. :) The PR was made a bit on the run, hence not being able to test that.

ozone/core/calculated/baselines.py
sorinstelian: In the end, I renamed and lru_cached the method.

PR #1425, opened 2019-10-28
Normalize to bool

mgax:
When importing xlsx files generated by our export code, these values arrive as integers (`1` or `0`). When the model instance gets created, for some esoteric reason, some `1`s get converted to `True` and some remain as `1`. (Of course, when the model is loaded from DB, they all come back as `True`, but during submission import, the model instance is newly created in memory.) At some point [this](https://github.com/eaudeweb/ozone/blob/d164aa8fb6f52b48984cc32752e432b2acedc80c/ozone/core/models/reporting.py#L1489) gets called and `is True` diligently checks for precisely the value `True`, rejecting poor `1`, and causing a bunch of `ProdCons` records to be skipped, and much grief on my end.


ozone/core/management/commands/import_submissions.py
```diff
@@ -711,16 +711,16 @@ def get_data(self, row, party, period):
...
+                "flag_has_reported_a1": bool(overall["AI_ComplRep"]),
+                "flag_has_reported_a2": bool(overall["AII_ComplRep"]),
+                "flag_has_reported_b1": bool(overall["BI_ComplRep"]),
+                "flag_has_reported_b2": bool(overall["BII_ComplRep"]),
+                "flag_has_reported_b3": bool(overall["BIII_ComplRep"]),
+                "flag_has_reported_c1": bool(overall["CI_ComplRep"]),
+                "flag_has_reported_c2": bool(overall["CII_ComplRep"]),
+                "flag_has_reported_c3": bool(overall["CIII_ComplRep"]),
+                "flag_has_reported_e": bool(overall["EI_ComplRep"]),
+                "flag_has_reported_f": bool(overall["F_ComplRep"]),
```
alexkiro: FWIW, this will only work if the values are always integers, while previously the DB will have automatically cast the string `"0"` to False, and this will treat it as True. 

Not sure if the excel file will always have an integer here, as I know they can be finicky :)

alexkiro: Looks good to me! :+1:

Issue #1339, opened 2019-09-30
Baseline HFC data report (raw data)


melish:
Similar to the raw data report, but 
- using predefined periods (see #1338): 2011-2013 and 1989 for NA5 parties, 2020-2022+2009-2010 for NA5G1, 2024-2026 + 2009-2010 for NA5G2
- showing only annex group F for the base periods (including blends containing annex F substances) and annex groups A/I and C/I for the corresponding baseline periods (1989 and respectively 2009, 2010)

See email from 30/07 and sample reports

PR #1426, opened 2019-10-30
Baseline HFC data report

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1339

* [x] Refactor the art7 raw data report so that it can receive custom querysets
* [x] Generate the report
* [x] Filter the substances to only show annex F


melish:
@mgax I spotted two issues in this report:
- please reorder the years in descending order (2013, 2012, 2011, 1989)
- please exclude the blend components from the substances (e.g. there should be only one HFC-32 for Australia, 2011, imports)

mgax:
@melish should be fixed.

PR #1433, opened 2019-11-06
Fixed A/I and C/I base periods for A5 parties, fixes #1432

### LINKED ISSUE #1432: HFC baseline report for A5 parties
Related to #1338 and #1417
I just realized that 1989 is hardcoded in `hfc_baseline.py`. This period only applies to NA5 parties, for A5 there should be 2009 and 2010 instead.

melish:


sorinstelian:
:+1:

Issue #1427, opened 2019-11-04
Production and consumption reports (variants)


melish:
All report variants show the comparison of selected period with baseline:

Annex 4a - summary for all parties
Annex 4b - summary for non A5 parties
Annex 4c - summary for A5 parties
Annex 6a - summary by region

PR #1438, opened 2019-11-11
Prodcons summary report

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1427.

* Contains some more refactoring of the prodcons report
* [x] Aggregation by region
  * [x] Aggregate values for prod, cons, and baselines, grouped by region
  * [x] Population aggregated by region
  * [x] Aggregate values for prod, cons, baselines over all the parties
* [x] A5 and NA5 parties aggregated
* [x] A5 and NA5 parties not aggregated


melish:
TODO: test EU population and per capita consumption (which normally should be calculated vs the entire population of the EU rather than individual parties who have/have not reported).

mgax:
> test EU population and per capita consumption

I've checked: the current formula result matches with the sample report. If I count the EU population as a whole, in the report by region, EU contains countries both in `WEUR` and `EEUR`, so the math gets complicated. I'd leave it as-is for now.

melish:
The %Chng column does not appear anymore in the new version of the old ProdCons vs baseline report

melish:
@mgax See this report for Australia, period C1999.
I think Consumption should appear as N.R. instead of blank. Why is it N.R. only for Production, but not Consumption?

Issue #1411, opened 2019-10-23
Laboratory and Analytical Uses report


melish:
rpt-22-lab-uses

This is identical to the final section of the Art 7 raw report, but without the rest.
See section_labuses.py

PR #1440, opened 2019-11-19
Export lab uses

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1411


ozone/core/api/export_pdf/export.py
```diff
@@ -93,6 +93,24 @@ def export_baseline_hfc_raw(parties):
     return buff
 
 
+def export_labuse(periods):
+
+    buff, doc = get_doc_template(landscape=True)
```
melish: use landscape=False pls

PR #1448, opened 2019-11-20
Invalidate website cache for all country profile models

melish:


sorinstelian:
@mgax : actually this PR is extending upon my approach - the reason I decided to use custom signals was to be able to use `send_robust` and avoid not saving things in the DB just because an external cache could not be invalidated.

ozone/core/signals.py
```diff
@@ -21,17 +40,31 @@ def clear_aggregation_cache(sender, instance, **kwargs):
...
+def clear_country_profile_cache(sender, instance, **kwargs):
     """
-    Handler for the clear_cache signal
+    Clear the cache of the main website
     """
-    invalidate_focal_point_cache(instance)
+    try:
+        invalidate_party_cache(instance.party_id)
+    except Exception as e:
+        logger.error(
```
mgax: You could use `logger.exception("Error while invalidating country profile cache.")` which will log the traceback.

By the way, extra arguments to `logger.error` only get included if they are interpolated in the message as e.g. `%s`.

Issue #1143, opened 2019-07-09
Production, Import and Export  - Comparison with Base Year


melish:
See `rpt-10-prod-imp-exp-cpmare-w-base.pdf`

melish:
How to get/store imports and exports for baseline year?

melish:
A: probably define new types of baseline records

melish:
On a second thought, it's not worth pre-computing baselines for imports and exports, as they are only used in this report.
Let's try calculating them on the fly. The logic is similar to consumption baselines from baselines.py, but using `import_new` and `export_new` from ProdCons.

A/I and A/II, NA5: value from 1986
A/I and A/II, A5: average of 1995-1997
B/I, B/II, B/III, NA5: value from 1989
B/I, B/II, B/III, A5: average of 1998-2000
C/I, NA5: exports/imports of C/I in 1989 + 2.8% of exports/imports of A/I in 1989
C/I, A5: average of 2009-2010
C/II, C/III: no baseline

E/I, NA5: value from 1991
E/I, A5: average of 1995-1998

F, NA5 (both NA5G1 and NA5G2): average for 2011-2013 + 15% of C/I baseline (see above)
F, A5G1: average of 2020-2022 + 65% of C/I baseline (average of C/I for 2009-2010)
F, A5G2: average of 2024-2026 + 65% of C/I baseline (average of C/I for 2009-2010)


* when computing

PR #1449, opened 2019-11-20
Report for "production, import and export"

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1143


melish:
@mgax Please check roundings - see 2018 report
http://prntscr.com/q5xxic

melish:
The data should also be filtered by the selected parties, if any. In general, party users should not see unaggregated data of other parties

mgax:
> The data should also be filtered by the selected parties, if any.

Right, fixed.

melish:
- [ ] Values are still not ok, now all values are rounded to the nearest int, which is not correct.
The number of decimals should depend on the annex group and period, see cached property ProdCons.decimals. In general, there should be 2 decimals in most cases, 1 decimal for C/I before 2010 and zero decimals for annex F.
- [ ] Please add group description after the group names (e.g. Hydrochlorofluorocarbons (HCFCs)). Resize the columns (again!) if needed.

mgax:
@melish I've fixed the missing baselines (they were not being calculated, forgot to copy/paste the formula) and the rounding decimal places. Please check again.

Issue #1414, opened 2019-10-23
Baseline prod/cons report for NA5 parties (C/I)


melish:
rpt-25-na5-1989-baseline-hcfc

Similar to #1338, but using a vertical layout, because the formula is simpler.
Maybe we can reuse the same layout, though

mgax:
The sample report `rpt-25-na5-1989-baseline-hcfc` contains values from the AI NA5Prod and NA5Cons baselines even though the columns are marked "CI baseline". The PR uses values from "CI" though.

Issue #1412, opened 2019-10-23
Baseline prod/cons report for A5 parties


melish:
rpt-13-baseline-cons-a5
rpt-14-baseline-prod-a5

These two reports show the production|consumption baseline for each party and the production|consumption for reference years for the "easy" annex groups: A/I, A/II, B/I, B/II, B/III, C/I, E/I

Reference years - currently hardcoded in various places. I think it's time to find a better place for them.
See also #1143
A/x: 1995-1997
B/x: 1998-2000
C/I: 2009-2010
E/I: 1995-1998

Issue #1453, opened 2019-11-25
Allow generating reports with no party or period selected


mgax:
[These reports](https://github.com/eaudeweb/ozone/issues/1412) can be generated with no parameters, and it will apply for all Art5 parties. Or if you select a party then only that party will be included in the report. But the UI gives an error, "At least one party or one period is required", if no selection is made, for either party or period.

<img width="786" alt="Screenshot 2019-11-25 at 14 34 15" src="https://user-images.githubusercontent.com/27617/69540977-b5eae980-0f90-11ea-86c7-a11537b1fb5d.png">

melish:
To discuss after #1447

PR #1454, opened 2019-11-25
Baseline prod/cons report for A5 parties

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1412, fixes https://github.com/eaudeweb/ozone/issues/1414.

Until https://github.com/eaudeweb/ozone/issues/1453 is fixed, select a period (it's ignored) and no party, and the report will be generated for all art5 parties.


melish:
Please check these small issues:
- parties having 0 consumption/production should also be included in the tables
- Micronesia should be included in the Consumption report - annex E/I (has consumption 0.1 for 1996), but it's missing
- there are some small differences in the subtotals compared to the sample report - possibly related to the previous issue

mgax:
>     * Micronesia should be included in the Consumption report - annex E/I (has consumption 0.1 for 1996), but it's missing

That's because it has no baseline in the database; how should I handle this case?

melish:
Per capita consumption subtotal should not be calculated as the sum of all values above, but total_cons/total_pop.

PR #1456, opened 2019-11-26
Auto-saving is_article5 and is_eu_member for MT aggregations

### LINKED ISSUE #1455: is_article5 and is_eu_member flags not populated on staging/prod
This is due to the fact that ProdConsMT objects do not have these flags auto-saved.

sorinstelian:
Also adding mgmt command for migrating existing values.

Closes #1455


ozone/core/management/commands/update_aggregation_flags.py
```diff
@@ -0,0 +1,73 @@
...
+
+        for ph in PartyHistory.objects.all():
+            if options['odp'] is True:
+                qs = ProdCons.objects.filter(
+                    party=ph.party, reporting_period=ph.reporting_period
+                )
+                qs.update(
+                    is_article5=ph.is_article5, is_eu_member=ph.is_eu_member
+                )
+                updated_odp += qs.count()
```
mgax: You can use the return value of [update](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#update), it's the count.

Issue #1457, opened 2019-11-26
Import and Export of New and Recovered ODS (summary report)


melish:
rpt-anx08.pdf

similar to #1145, but grouped by A5 status

Issue #1413, opened 2019-10-23
Used and Recycled ODSs Imported and Exported report


melish:
rpt-23-used-recycled-imp-exp or rpt-anx09.pdf (Annex IX)

similar to #1145, but detailed by substance

PR #1458, opened 2019-11-26
Recovered imports and exports by substance

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1413, fixes https://github.com/eaudeweb/ozone/issues/1457


melish:
Some subtotals are missing for `Recovered ODSs Imported and Exported by the Parties` (for each party, for all Article 5 / Non-Article 5 and the overall total for the selected year)

mgax:
>     * There are some rounding differences in the aggregated report.

Is it just that different value or are there others? Maybe some party really did update their reported value.

mgax:
@melish all fixed. But a lot of values are different from the example. I'm printing values from the `Article7Import` and `Article7Export` tables; should they be converted to CO2-equivalent tonnes maybe?

Issue #1447, opened 2019-11-20
Refactor reports


melish:
Now that the number of reports is increasing, it would be good to refactor the entrypoints, parameters and how they are defined.

PR #1459, opened 2019-11-27
Refactor reports

mgax:
Fixes https://github.com/eaudeweb/ozone/issues/1447

~Currently this PR is comparing to `merge-all-reports`, but it will be against `develop`, once the 3 remaining report PRs are merged.~

It adds some tests that simply check that each report can be generated as a valid PDF, no checks are made on the content. The plan is to make each report a subclass of `ozone.core.api.export_pdf.util.Report` and generate the API from the information in those classes.

~So far, a single report is thusly refactored, `ProdImpExpReport`. I'd appreciate some feedback on the general approach :)~

All reports are now generated by classes that descend from `Report` and they share a common page layout and http response logic. The report subclasses are responsible for defining their name, what parameters they expect, and to generate a list of reportlab flowables.

Three reports (raf, art7raw and prodcons) can be generated for a given submission. They descend from the `ReportForSubmission` class that abstracts the code for this common case.

The classes are now registered in the list `ozone.core.api.export_pdf.reports.registry` and the api generates its report listing from there.


ozone/core/api/export_pdf/__init__.py
```diff
@@ -7,7 +7,6 @@
     export_prodcons_a5_summary,
```
melish: I may feel better if you rename export_* methods. Sometimes it's confusing (for reports about export and import). 
I think names like `def impexp_new_rec(periods, parties)` etc. should be clear enough.
This can be done later, too

ozone/core/api/export_pdf/reports/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .prodcons.prod_imp_exp import ProdImpExpReport
+
+registry = [
+    ProdImpExpReport,
```
melish: Will all reports be added here?

ozone/core/api/export_pdf/reports/prodcons/prod_imp_exp.py
```diff
@@ -221,16 +218,22 @@ def render_header(period):
...
-def get_prod_imp_exp_flowables(periods, parties):
-    all_groups = data.get_all_groups()
-    groups_description = list(render.get_groups_description(all_groups))
+class ProdImpExpReport(Report):
+
+    name = "prod_imp_exp"
+    has_party_param = True
+    has_period_param = True
+
+    def get_flowables(self):
```
melish: nice that we will have `get_flowables` everywhere

ozone/core/api/export_pdf/util.py
```diff
@@ -224,7 +241,7 @@ def to_precision(nr, decimals):
         s_nr = get_big_float(nr)
 
         # Getting the first non-zero digitindex
-        p = re.compile('(?=\d)(?=[^0])')
+        p = re.compile(r'(?=\d)(?=[^0])')
```
melish: We will remove to_precision at some point, it shouldn't be used. For now it's only used by the "HAT" PDF output, which is not fully implemented.

ozone/core/api/views.py
```diff
@@ -2464,215 +2468,169 @@ class ReportsViewSet(viewsets.ViewSet):
...
-        params = "%s_%s" % (
-            "_".join(p.abbr for p in parties),
-            "_".join(p.name for p in periods),
-        )
-        return self._response_pdf(
-            f'prod_imp_exp{params}',
-            export_prod_imp_exp(periods=periods, parties=parties)
-        )
+        from ozone.core.api.export_pdf.reports import ProdImpExpReport
+        return ProdImpExpReport.from_request(request).render_to_response()
```
melish: like

ozone/core/api/export_pdf/reports/__init__.py
mgax: Yep, that's the plan.

ozone/core/api/export_pdf/util.py
mgax: Sure, whatever, but this line generated a flake8 warning, because `\d` is not a valid escape, so it's left as `\d` in the actual string. I converted it to a raw string which, in this case, has the same meaning.

ozone/core/api/export_pdf/__init__.py
mgax: All these functions are going away, all they do is duplicate the code, which is moving to [this method](https://github.com/eaudeweb/ozone/blob/8976554d2b202b95e84859b07954874ca07be04e/ozone/core/api/export_pdf/util.py#L630-L640).

PR #1486, opened 2019-12-09
Filter out EU or EU members when aggregating certain fields

### LINKED ISSUE #1479: Do not count EU when aggregating data for Production/Destruction
Calculated Production is already set to None for the European Union, so only the Destruction part should be implemented.

### LINKED ISSUE #1485: Aggregations -  ignore member states data for consumption
When aggregating by party, all consumption-related data (imports, exports, consumption) for EU members should be ignored.

sorinstelian:
Closes #1479, #1485.

### DIFF
--- a/ozone/core/api/views.py
+++ b/ozone/core/api/views.py
@@ -812,7 +812,7 @@ class AggregationViewSet(viewsets.ReadOnlyModelViewSet):
         for value in all_values:
             values_dict[value['reporting_period']].append(value)
 
-        if 'party' in aggregates:
+        if aggregates and 'party' in aggregates:
             exclude_eu_and_members = True
         else:
             exclude_eu_and_members = False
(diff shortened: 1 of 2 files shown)


ozone/core/api/views.py
```diff
@@ -773,6 +811,13 @@ def list_aggregated_data(
         for value in all_values:
             values_dict[value['reporting_period']].append(value)
 
+        if 'party' in aggregates:
+            exclude_eu_and_members = True
+            eu_party_id = Party.objects.get(abbr='EU').id
```
melish: Try caching this id (it's probably needed elsewhere, too)

sorinstelian: Added as a cached function (to avoid short object lifetime making caching useless) in `core/models/party.py`; could be of use elsewhere too.

Issue #1401, opened 2019-10-18
View/Edit mode for submission forms


melish:
Submissions can be opened from the dashboard using the "View" or "Edit" button.
In some cases, even when "View" is used, the submission is still editable (e.g. for adding comments, uploading files, changing flags, etc.).
To prevent user mistakes, we were told to add a functionality similar to #1176:

- instead of the Save button show an "Edit" button, which will switch to edit mode when pressed and will change its label to "Save"
- if too complicated, there is no need to change the behaviour of other fields (like disabling them until "Edit mode" is on)

melish:
As part of this task, we should re-visit the permissions of the secretariat and fix some known bugs like:

- the questionnaire checkboxes are editable, even though the backend does not allow changes
- the "Checked blank flags" are disabled while state is "Submitted", but editable while "Processing"
- OzSec wants to be able to make changes in the submission info fields while state is "Processing" (to be clarified)

PR #1522, opened 2019-12-20
View/Edit mode for submissions, refs #1401

melish:
This is replaced by pull request #1580


melish:
Here are a few initial bugs. I've only tested a new Art7 submission:

- Fields from `Submission information` (including the flags) are not being saved.
- Checkboxes in the Art7 Questionnaire tab are still editable when Edit mode is not active
- `Add facility` button from tab `Emisssions` is still enabled when Edit mode is not active
- When creating a new submission, the Edit mode should be active by default
- Once `Edit` is pressed, it should remain in the edit mode (don't revert to view mode after Save)

- Please add new tests for all the issues above

melish:
@razvanMiu Please rebase to develop and fix conflicts

melish:
Replaced by pull request #1580. Keeping this for a while for traceability reasons

PR #1565, opened 2020-03-04
Adding global cache invalidation for TEAP, IMPCOM and licensing systems

sorinstelian:
Solves backend part for https://helpdesk.eaudeweb.ro/issues/7981.

**Pending**: Not sure which models should trigger cache invalidation with the `status_of_ratification` param. PartyRatification?


melish:
> **Pending**: Not sure which models should trigger cache invalidation with the `status_of_ratification` param. PartyRatification?

Yes, PartyRatification. And changes to party ratification models should also invalidate the LicensingSystems

PR #1570, opened 2020-03-19
Allowing pre-fill for Article 9 and Transfer admin form

sorinstelian:


ozone/core/models/country_profile.py
```diff
@@ -170,10 +186,36 @@ class OtherCountryProfileData(models.Model):
...
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super().save(*args, **kwargs)
+
     class Meta:
         db_table = "other_country_profile_data"
         verbose_name_plural = "other country profile data"
         ordering = ("party__name", "-reporting_period__end_date")
+        unique_together = ('party', 'reporting_period', 'obligation')
```
melish: This should not be here

PR #1579, opened 2020-03-26
Edit mode v2, fixes #1401

### LINKED ISSUE #1401: View/Edit mode for submission forms

melish:
Rebased pull request #1522


melish:
Invalid

PR #1581, opened 2020-04-01
8026 - PDF diff between current and previous version of a submission

sorinstelian:
Can be used with:
http://<host>:<port>/api/submissions/<submission_pk>/export_diff_pdf/

Closes https://helpdesk.eaudeweb.ro/issues/8026

### DIFF
--- a/ozone/core/models/reporting.py
+++ b/ozone/core/models/reporting.py
@@ -1421,6 +1421,17 @@ class Submission(models.Model):
             obligation=self.obligation,
         ).prefetch_related('created_by')
 
+    def get_previous_version(self):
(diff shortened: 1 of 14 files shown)


ozone/core/models/reporting.py
```diff
@@ -1421,6 +1421,17 @@ def versions(self):
             obligation=self.obligation,
         ).prefetch_related('created_by')
 
+    def get_previous_version(self):
+        # TODO: implement this properly - maybe based on HistoricalSubmission
+        # or just on latest() on the superseded versions queryset
+        if self.version == 1:
+            return None
+        return Submission.objects.filter(
```
melish, line 1429: Will this always work?

PR #1585, opened 2020-04-30
Adding CAP users with special permissions

sorinstelian:
Closes https://helpdesk.eaudeweb.ro/issues/8226 https://helpdesk.eaudeweb.ro/issues/8051

For https://helpdesk.eaudeweb.ro/issues/8051:
All country profile endpoints allow read-only access for any authenticated user; special permission needed to be added only for aggregation ODP & MT endpoints (inc. destruction).


ozone/core/api/views.py
```diff
@@ -422,7 +434,7 @@ class GroupSubstanceViewSet(ReadOnlyMixin, viewsets.ModelViewSet):
 
     queryset = Group.objects.all()
     serializer_class = GroupSubstanceSerializer
-    # Allows unauthenticatd GET requests
+    # Allows unauthenticated GET requests
```
melish, line 437: Is 'unauthenticated' the proper word? Or rather, 'unprivileged'?

ozone/core/api/views.py
```diff
@@ -1653,7 +1676,8 @@ class SubmissionInfoViewSet(viewsets.ModelViewSet):
     obligation_types = None
     serializer_class = SubmissionInfoSerializer
     permission_classes = (
-        IsAuthenticated, IsSecretariatOrSamePartySubmissionRelated,
+        IsAuthenticated,
+        IsSecretariatOrSamePartySubmissionRelated,
```
melish, line 1680: CAP users should be able to see the SubmissionInfo, too

ozone/core/api/views.py
```diff
@@ -753,7 +765,8 @@ class AggregationViewSet(viewsets.ReadOnlyModelViewSet):
     )
     ordering = ("-reporting_period__start_date", "party", "group")
     permission_classes = (
-        IsAuthenticated, IsSecretariatOrSamePartyAggregation,
+        IsAuthenticated
```
melish, line 768: File "/var/local/ozone/ozone/core/api/views.py", line 769, in AggregationViewSet
    & (IsSecretariatOrSamePartyAggregation | IsMobileApp)
TypeError: unsupported operand type(s) for |: 'type' and 'type'

ozone/core/api/views.py
sorinstelian: Yes, they will be able to see it due to Submission.has_read_rights_for_party(user) being called.

ozone/core/api/views.py
sorinstelian: Yes, it allows read-only access to group-substance data even for unauthenticated users. I think this is currently being used by the Drupal site.

PR #1591, opened 2020-05-12
Using backend permissions for frontend checks

sorinstelian:
From the backendside, only the relevant value for each dictionary key is returned when GET-ting `http://localhost:8000/api/submissions/<pk>/`. E.g.:
```
    "permission_matrix": {
        "reporting_channel": true,
        "submission_format": true,
        "reporting_officer": true,
        "submitted_at": true,
        "other_submission_info_data": true,
        "flag_provisional": false,
        "flag_blank": true,
        "flag_annex_group": true,
        "files": true,
        "questionnaire": true,
        "substance_data": true,
        "remarks_party": true,
        "remarks_secretariat": false
    }
```


sorinstelian:
LGTM

sorinstelian:
@razvanMiu : I did some testing from the frontend and it looks like:
- with OS user, on Party-created submission in Finalized state, I can edit the "provisional" flag (which should be RO) and the "has_blanks" and "confirmed_blanks" flags (which should also be RO due to `flag_blank` being True).

melish:
Looks ok, but once we switch to the new way of checking permissions, we should remove previous properties like "response.data.can_change_reporting_channel", "can_upload_files", etc , just to prevent other unexpected behaviour spreaded around the frontend code

PR #1594, opened 2020-05-19
Filter parties by party groups (CAP users), refs #8226

melish:


ozone/core/api/export_pdf/util.py
```diff
@@ -630,15 +630,22 @@ def get_doc_template(landscape=False):
...
         qs = Party.get_main_parties()
-        if parties:
-            qs = qs.filter(pk__in=parties)
     else:
+        related_parties = [
+            request.user.party_id
+        ] if request.user.party_id else []
+        if request.user.party_group:
+            related_parties += [
+                party.id for party in request.user.party_group.parties.all()
```
sorinstelian, line 641: Probably no real need to optimize here, but I'd use `values_list('id', flat=True)` instead of iterating over all parties.

PR #1600, opened 2020-05-21
More fixes for "diff" PDF report

sorinstelian:


sorinstelian:
Making this "ready for review".
There seem to be a few small issues remaining:
- "Subtotal" rows are included in the Exports diff section (not sure if this is a problem)
- A "Laboratory and analytical uses under the global exemption (metric tonnes)" section is added, without presenting any obvious diff (though this certainly exists). Should this be removed entirely?

PR #1607, opened 2020-05-28
8307 date received on RAF submissions

melish:


ozone/core/api/export_pdf/util.py
```diff
@@ -536,26 +537,70 @@ def get_submission_dates(submission):
...
+        return (
+            Paragraph('%s: %s%s' % (
+                _('Date received'),
+                format_date(date_received),
+                extra_text,
+            ), style=no_spacing_style),
+            p_l(''),
+        )
+    else:
+        return (None,)
```
sorinstelian, line 603: In my (very limited) experience with reportlab, having None in flowables sometimes led to hard to debug crashes. Why not return an empty tuple?

melish: The return value is wrapped in a tuple because it's going to be added with other tuples in method export_info. Reportlab actually receives the None (without the tuple) as a "main" element and seems to be working well (completely skipped when rendering).
See here https://github.com/eaudeweb/ozone/pull/1607/files#diff-51d737393c6952ba9f14789564be0edbL205

melish: On a second though, empty tuple would be fine, too, if not better.

PR #1625, opened 2020-07-15
Fixing errors in questionnaire diff; improving info on compared submissions

sorinstelian:
What has been done so far:
- Avoid displaying questionnaire diff for identical submissions
- Fixing questionnaire diff when no answers selected
- Displaying versions being compared
- Displaying party-period-version when no previous submission found

Closes #8373

### DIFF
--- a/ozone/core/api/export_pdf/art7/section_info.py
+++ b/ozone/core/api/export_pdf/art7/section_info.py
@@ -89,6 +89,9 @@ def get_submission_info(info):
 
 def get_questionnaire_table_contents(submission):
     def _yn(condition):
+        # Questionnaire flags are nullable booleans
+        if condition is None:
+            return '-'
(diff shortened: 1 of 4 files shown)


ozone/core/api/export_pdf/art7/section_info.py
```diff
@@ -87,8 +106,55 @@ def get_submission_info(info):
...
+        "%s: %s (%s)" % (
+            _('Name of reporting officer'), reporter, previous_reporter
+        ),
+        style=no_spacing_style
+    )
+
+    address_country = None
+    previous_address_country = None
+    if info.country and info.country.name != info.submission.party.name:
+        address_country = info.country_name
```
melish, line 131: I believe info.country_name is not a real thing

PR #1639, opened 2021-01-06
[Bug #8421] Saving "Other" categories in RAF (essencrit) submission

razvanMiu:
* Recursively computed error message
* True delete of use_categories
* Don't save use_category if quantity == 0

### DIFF
--- a/frontend/src/components/raf/Save.vue
+++ b/frontend/src/components/raf/Save.vue
@@ -5,13 +5,16 @@ export default {
   mixins: [SaveMixin],
   methods: {
     getData(tab) {
-      return {
-        ...tab,
-        form_fields: tab.form_fields.map(form => ({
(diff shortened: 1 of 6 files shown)


frontend/src/components/raf/Save.vue
```diff
@@ -3,6 +3,21 @@ import SaveMixin from '@/components/common/mixins/SaveMixin'
 
 export default {
   mixins: [SaveMixin],
-  methods: {}
+  methods: {
+    getData(tab) {
+      return {
+        ...tab,
+        form_fields: tab.form_fields.map(form => ({
```
melish, line 10: When a field in SubmissionInfo tab is changed, this line triggers an exception on save, because tab.form_fields.map is undefined.
Please fix

PR #1646, opened 2021-04-06
Small patch

razvanMiu:


razvanMiu:
This also disable the filtering on version column related to [Bug #9269].
TODO: make sure that api support filtering for revision (version) field.

